PR #80, opened 2015-01-08
Update 11-haxelib.tex

yar3333:
haxelib.json - about classPath


yar3333:
haxelib.json - about classPath

Simn:
I actually have no idea what classPath is, so I'll just trust you on this.

PR #84, opened 2015-01-27
Added General metadata table

markknol:
These values are grabbed from the --help-metas, formatted as a table, I added some reference links. I added @:enum manually since that one was missing.

I added this, because I look into http://old.haxe.org/manual/tips_and_tricks quite a lot.

Some stuff is empty, I have no idea if how up-to-date it is.


Simn:
Thanks for your work, but the markdown is generated from the LaTeX sources so it would be overwritten on the next export.

I might hack something into the markdown converter to execute Haxe and grab the output automatically.

PR #87, opened 2015-01-29

markknol:
Missing code blocks, added link to conditional compilation page.


Simn:
I think at some point I wanted to use `\expr{}` only for actual Haxe expressions so we could add some syntax highlighting later. There was some other command to use for general monospace. I didn't really go through with that though, so I'm fine with merging. Thanks!

markknol:
Ok good to know, so what should i use for non-code things? Like commandline/filename expressions and stuff?

Simn:
I forget... :D

markknol:
I think it is `/ic{}` ?

Simn:
That might be it! Though I have no idea what that stands for.

PR #90, opened 2015-01-30
Added Lambda

markknol:
I took some of the old manual, but rewrote to a compacter story with 2 examples and a clear list what's in it.

### DIFF
--- a/10-std.tex
+++ b/10-std.tex
@@ -295,9 +295,9 @@ It is common to see \Fullref{lf-static-extension} used with the math library.  T
 \section{Lambda}
 \label{std-Lambda}
 
-\define{Lambda}{define-lambda}{Lambda is a functional language concept within Haxe that allows you to apply a function to a set or \tref{custom iterators}{lf-iterators}. The Lambda class is a collection of functional methods in order to use functional-style programming with Haxe.}
+\define{Lambda}{define-lambda}{Lambda is a functional language concept within Haxe that allows you to apply a function to a list or \tref{iterators}{lf-iterators}. The Lambda class is a collection of functional methods in order to use functional-style programming with Haxe.}
 
-It is ideally used with \expr{using Lambda} (read \tref{Static Extension}{lf-static-extension}) and then acts as an extension to \type{Iterable} types. 
+It is ideally used with \expr{using Lambda} (see \tref{Static Extension}{lf-static-extension}) and then acts as an extension to \type{Iterable} types. 
 
 On static platforms, working with the \type{Iterable} structure might be slower than performing the operations directly on known types, such as \type{Array} and \type{List}.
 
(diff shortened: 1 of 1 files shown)


10-std.tex
```diff
@@ -295,7 +295,79 @@ \subsection{Extensions}
...
+	\item[\expr{Lambda.find}] Find first element of given search function.
+	\item[\expr{Lambda.foreach}] Determine if every element satisfies a criteria.
+	\item[\expr{Lambda.iter}] Call a function for each element.
+	\item[\expr{Lambda.concat}] Merg two Iterables, returning a new List.
+	\item[\expr{Lambda.filter}] Find the elements that satisfy a criteria, returning a new List.
+	\item[\expr{Lambda.map}, \expr{Lambda.mapi}] Apply a conversion to each element, returning a new List.
+	\item[\expr{Lambda.fold}] Functional fold, which is also known as reduce, accumulate, compress or inject.
+\end{description}
+
+This example demonstrates the Lamda filter and map on a set of strings:
```
Simn: typo Lamda

10-std.tex
```diff
@@ -295,7 +295,79 @@ \subsection{Extensions}
 \section{Lambda}
 \label{std-Lambda}
 
+\define{Lambda}{define-lambda}{Lambda is a functional language concept within Haxe that allows you to apply a function to a set or \tref{custom iterators}{lf-iterators}. The Lambda class is a collection of functional methods in order to use functional-style programming with Haxe.}
```
Simn: "to a set" What's a set? We don't really use that term.

10-std.tex
```diff
@@ -295,7 +295,79 @@ \subsection{Extensions}
...
 
+It is ideally used with \expr{using Lambda} (read \tref{Static Extension}{lf-static-extension}) and then acts as an extension to \type{Iterable} types.
```
Simn: I think it's more common to use "see" instead of "read" here.

10-std.tex
```diff
@@ -295,7 +295,79 @@ \subsection{Extensions}
 \section{Lambda}
 \label{std-Lambda}
 
+\define{Lambda}{define-lambda}{Lambda is a functional language concept within Haxe that allows you to apply a function to a set or \tref{custom iterators}{lf-iterators}. The Lambda class is a collection of functional methods in order to use functional-style programming with Haxe.}
 
+It is ideally used with \expr{using Lambda} (read \tref{Static Extension}{lf-static-extension}) and then acts as an extension to \type{Iterable} types. 
+
+On static platforms, working with the \type{Iterable} structure might be slower than performing the operations directly on known types, such as \type{Array} and \type{List}.
```
Simn: I tried to talk about "targets" instead of "platforms" throughout the manual because I think it is more accurate.

10-std.tex
```diff
@@ -295,7 +295,79 @@ \subsection{Extensions}
...
+
+\paragraph{Lambda Functions}
+The Lambda class allows us to operate on an entire \type{Iterable} at once.
+This is often preferable to looping routines since it is less error prone and easier to read. 
+For convenience, the \type{Array} and \type{List} class contains some of the frequently used methods from the Lambda class.
```
Simn: There are a lot of newlines here. Does this look alright in the output?

10-std.tex
```diff
@@ -295,7 +295,79 @@ \subsection{Extensions}
...
+It is helpful to look at an example. The exists function is specified as:
+
+\begin{lstlisting}
+static function exists<A>( it : Iterable<A>, f : A -> Bool ) : Bool
+\end{lstlisting}
+
+Most Lambda functions are called in similar ways. The first argument for all of the Lambda functions is the \type{Iterable} on which to operate. Many also take a function as an argument.
+
+\begin{description}
+	\item[\expr{Lambda.array}, \expr{Lambda.list}] Convert Iterable to \type{Array} or \type{List}. It returns a copy if it's already of that type.
```
Simn: That's not correct, it always creates a new instance.

10-std.tex
```diff
@@ -295,7 +295,79 @@ \subsection{Extensions}
...
+	\item[\expr{Lambda.count}] Count the number of elements.  If the Iterable is a \type{Array} or \type{List} it is faster to use its length property.
+	\item[\expr{Lambda.empty}] Determine if the Iterable is empty. For all Iterables it is best to use the this function; it's also faster than compare the length (or result of Lambda.count) to zero.
```
Simn: I highly doubt that checking for `it.iterator().hasNext()` is faster than comparing an integer. Where does that information come from?

10-std.tex
```diff
@@ -295,7 +295,79 @@ \subsection{Extensions}
...
+	\item[\expr{Lambda.has}] Determine if the specified element is in the Iterable.
+	\item[\expr{Lambda.exists}] Determine if any element satisfies a criteria.
```
Simn: Something about that "any" bothers me. I think we should be more precise here.

10-std.tex
```diff
@@ -295,7 +295,79 @@ \subsection{Extensions}
...
+	\item[\expr{Lambda.indexOf}] Find out the index of the specified element.
+	\item[\expr{Lambda.find}] Find first element of given search function.
+	\item[\expr{Lambda.foreach}] Determine if every element satisfies a criteria.
+	\item[\expr{Lambda.iter}] Call a function for each element.
+	\item[\expr{Lambda.concat}] Merg two Iterables, returning a new List.
```
Simn: typo merg

10-std.tex
markknol: one newline means nothing, two means a new paragraph. so it is safe to use it. I found it a bit easier to read in my editor.

10-std.tex
markknol: It comes from the old documentation. I haven't tested it. I _think_ it should be faster than Lambda.count since that first iterates over the full list.

10-std.tex
```diff
@@ -295,7 +295,79 @@ \subsection{Extensions}
...
+\end{lstlisting}
+
+Most Lambda functions are called in similar ways. The first argument for all of the Lambda functions is the \type{Iterable} on which to operate. Many also take a function as an argument.
+
+\begin{description}
+	\item[\expr{Lambda.array}, \expr{Lambda.list}] Convert Iterable to \type{Array} or \type{List}. It always returns a new instance.
+	\item[\expr{Lambda.count}] Count the number of elements.  If the Iterable is a \type{Array} or \type{List} it is faster to use its length property.
+	\item[\expr{Lambda.empty}] Determine if the Iterable is empty. For all Iterables it is best to use the this function; it's also faster than compare the length (or result of Lambda.count) to zero.
+	\item[\expr{Lambda.has}] Determine if the specified element is in the Iterable.
+	\item[\expr{Lambda.exists}] Determine if given element satisfies a criteria.
```
Simn: That's not good either because the function doesn't receive an element. How about "Determine if criteria is satisfied by an element."?

markknol: Sounds good to me :smile:

Simn:
Thanks a lot!

PR #103, opened 2015-02-23
PHP target details

markknol:
Contains; Getting started with PHP.

### DIFF
--- a/12-target-details.tex
+++ b/12-target-details.tex
@@ -6,7 +6,7 @@
 \label{target-javascript}
 \state{NoContent}
 
-\subsection{Getting started with Javascript}
+\subsection{Getting started with Haxe/Javascript}
 \label{target-javascript-getting-started}
 
 Haxe can be a powerful tool for developing Javascript applications. Let's look at our first sample.
@@ -29,13 +29,13 @@ class Main {
 }
 \end{lstlisting}
 
-To compile, either run this from the command line:
+To compile, either run the following from the command line:
 
 \begin{lstlisting}
 haxe -js main-javascript.js -main Main -D js-flatten -dce full
 \end{lstlisting}
 
-.. or create and run (double-click) a file called \ic{compile.hxml}. In this example the hxml-file should be in the same directory as the example class.
+Another possibility is to create and run (double-click) a file called \ic{compile.hxml}. In this example the hxml-file should be in the same directory as the example class.
 
 \begin{lstlisting}
 -js main-javascript.js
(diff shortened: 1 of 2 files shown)


12-target-details.tex
```diff
@@ -301,6 +301,46 @@ \section{Neko}
...
+\begin{lstlisting}
+-php bin
+-main Main
+\end{lstlisting}
+
+The compiler outputs in the given \emph{bin}-folder, which contains the generated PHP classes that prints the traced message when you run it.
+
+\paragraph{Run the PHP}
+
+To run the PHP a web server is needed, such as \href{http://httpd.apache.org/}{Apache}. An easy-to-install Apache distribution for local development is \href{https://www.apachefriends.org/download.html}{XAMPP}.
```
Simn: That's overkill, all you need is a PHP interpreter. You can easily run PHP from command line (I do it all the time when testing).

12-target-details.tex
```diff
@@ -301,6 +301,46 @@ \section{Neko}
 
 \section{PHP}
 \label{target-php}
+\state{NoContent}
+
+\subsection{Getting started with PHP}
+\label{target-php-getting-started}
+
+Haxe can be a powerful tool for developing PHP applications. Let's look at our first example showing the toolchain.
```
Simn: This doesn't carry any value whatsoever.

12-target-details.tex
```diff
@@ -301,6 +301,46 @@ \section{Neko}
...
 \section{PHP}
 \label{target-php}
+\state{NoContent}
+
+\subsection{Getting started with PHP}
+\label{target-php-getting-started}
+
+Haxe can be a powerful tool for developing PHP applications. Let's look at our first example showing the toolchain. 
+
+Create a new folder and save this class as \ic{Main.hx}.
```
Simn: I'd remove the line above and start here with "To get started with Haxe/PHP, create a new folder..."

We might want to label the section "Getting started with Haxe/PHP" too (and do the same for other targets) to not confuse google.

12-target-details.tex
```diff
@@ -301,6 +301,46 @@ \section{Neko}
...
+
+\haxe{assets/HelloPHP.hx}
+
+To compile, either run this from the command line:
+
+\begin{lstlisting}
+haxe -php bin -main Main
+\end{lstlisting}
+
+.. or create and run (double-click) a file called \ic{compile.hxml}. In this example the hxml-file should be in the same directory as the example class.
```
Simn: Ellipsis in technical documents are not good, but having only two dots is even worse. ;)

Can we restructure this?

markknol: It is also used in the flash/javascript section, ill have to fix those too.

Issue #23, opened 2014-05-22
Compiler features - Completion


Simn:

markknol:
Ok we have it here http://haxe.org/manual/cr-completion.html, fixed by https://github.com/HaxeFoundation/HaxeManual/pull/89

What about the BOM?

Simn:
Indeed the BOM should at least be mentioned.

markknol:
I have no idea how to describe this, can you help?

Simn:
I guess we should just say that the reported offsets are byte position, not char position.

PR #106, opened 2015-02-25
Completion char positions

markknol:
Closes https://github.com/HaxeFoundation/HaxeManual/issues/23 
Is this enough info?


Simn:
Ah, it already mentioned that it's byte position. I suppose it's good to highlight that like you did.

PR #111, opened 2015-03-03
removed underscore + no dots on of end description

markknol:


Simn:
The dots should be fixed in the compiler...

Simn:
Reminder to self: the `@:deprecated` description is deprecated.

markknol:
The dots are already fixed in the compiler :wink:

Simn:
Do are we planning to keep the changes in the compiler in sync with the manual?

markknol:
I think half, since the description itself should be the same, but I now also have some links to other pages in the manual. (I think this cannot be automated)

PR #113, opened 2015-03-16
Chp11 Available commands to same verb tense

aszasz:
In the description of available commands in Haxelib, descriptions were sometimes in future tense, sometimes in present tense. I standardize it to present and third person.
The thought was to follow a narrative description: "The available comand DOES this" instead of: "Available command: DO this".


Simn:
Missed this one, thanks!

PR #116, opened 2015-03-20
Corrections to the serialization format section

jonasmalacofilho:
Corrections:
 - Incorrect prefix in the example for Date
 - Missing `:` before the number of arguments of a enum constructor

Other changes:
 - String caching being always enabled is probably not important to the specification of the format; mention it next to the object caching settings
 - Describe both types of cache references together
 - Point out that cache indices and enum constructor indices start from zero
 - Insert empty mboxes to prevent inlining of the first items when nesting description environments

### DIFF
--- a/10-std.tex
+++ b/10-std.tex
@@ -533,8 +533,8 @@ The result of the serialization (here stored in local variable \expr{s}) is a \t
 Serialization can be configured in two ways. For both a static variable can be set to influence all \type{haxe.Serializer} instances, and a member variable can be set to only influence a specific instance:
 
 \begin{description}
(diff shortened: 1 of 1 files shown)


jonasmalacofilho:
This is related to jonasmalacofilho/haxe@298b7072e9c941b30b8776a5076baa9b6eff94df, which might become a PR someday if Simn thinks it's worth it.

I'm not sure about those tests: even though the current tests miss bugs and possible format changes, these new tests are quite repetitive and I'm also not proud of only testing the decoder.

Simn:
Thanks a lot!

PR #118, opened 2015-03-23
Template spelling fixes

markknol:


10-std.tex
```diff
@@ -397,7 +397,7 @@ \section{Template}
 ::if isValid:: valid ::else:: invalid ::end::
 \end{lstlisting} 
 
-Operators can be used, but it doesn't deal with operator precedence. Therefore it is required to enclose each operation in parentheses \ic{()}. The current operators are allowed: \ic{+}, \ic{-}, \ic{*}, \ic{/}, \ic{>}, \ic{<},  \ic{>=}, \ic{<=}, \ic{==}, \ic{!=}, \ic{&&} and \ic{||}.
+Operators can be used but they doesn't deal with operator precedence. Therefore it is required to enclose each operation in parentheses \ic{()}. Currently, the following operators are allowed: \ic{+}, \ic{-}, \ic{*}, \ic{/}, \ic{>}, \ic{<},  \ic{>=}, \ic{<=}, \ic{==}, \ic{!=}, \ic{&&} and \ic{||}.
```
Simn: "they don't"

10-std.tex
```diff
@@ -442,14 +442,14 @@ \section{Template}
 The console will trace \ic{The users are Mark(30) John(45)}.
 
 \paragraph{Macros}
-To callback code while parts of the template are rendered, providing a \expr{macros} object to the argument of \expr{Template.execute}. The key will act the template variable name, the value refers to a callback function that should return a \type{String}. The first argument of this macro function is always a \expr{resolve()} method, followed by the given arguments. The resolve function can be called to retreive values from the Template context. If \expr{macros} has no such field, the result is unspecified.
+To callback code while parts of the template are being rendered, provide a \expr{macros} object to the argument of \expr{Template.execute}. The key will act as the template variable name, the value refers to a callback function that should return a \type{String}. The first argument of this macro function is always a \expr{resolve()} method, followed by the given arguments. The resolve function can be called to retrieve values from the template context. If \expr{macros} has no such field, the result is unspecified.
```
Simn: "callback" is not really a verb. "To call custom functions while..."

10-std.tex
```diff
@@ -442,14 +442,14 @@ \section{Template}
...
 
-The following example passes itself as macro function context, and executes \ic{display} from the template.
+The following example passes itself as macro function context and executes \ic{display} from the template.
 \haxe{assets/TemplateMacros.hx}
 The console will trace \ic{The results: Mark ran 3.5 kilometers in 15 minutes}.
 
 \paragraph{Globals}
-Use the \expr{Template.globals} object to store those values that should be applied across all Template instances. This has lower priority than the context argument of \expr{Template.execute}.
+Use the \expr{Template.globals} object to store the values that should be applied across all template instances. This has lower priority than the context argument of \expr{Template.execute}.
```
Simn: The upper-case Template was actually good here, but it should be haxe.Template to really get the point across.

Simn: Also I would lose the "the" that used to be "those" entirely. "object to store values that"

markknol:
done

10-std.tex
```diff
@@ -441,15 +441,15 @@ \section{Template}
...
 
-The following example passes itself as macro function context, and executes \ic{display} from the template.
+The following example passes itself as macro function context and executes \ic{display} from the template.
 \haxe{assets/TemplateMacros.hx}
 The console will trace \ic{The results: Mark ran 3.5 kilometers in 15 minutes}.
 
 \paragraph{Globals}
-Use the \expr{Template.globals} object to store those values that should be applied across all Template instances. This has lower priority than the context argument of \expr{Template.execute}.
+Use the \expr{Template.globals} object to store values that should be applied across all \expr{haxe.Template} instances. This has lower priority than the context argument of \expr{Template.execute}.
```
Simn: nitpick: use \type for types so we can at some point link them to their API documentation.

markknol: Ok, should this still be a full path?

Simn: It should be \type{haxe.Template}

Simn:
Thanks!

PR #121, opened 2015-03-24
Convert: \slash, \noindent and empty \mbox

### LINKED ISSUE #120: unsupported tags error
https://github.com/HaxeFoundation/HaxeManual/commit/b0cf090686e7bd53b37a8fb0420b7755592d3451 this commit contains a `\slash`, `\noindent`, `\mbox`, which are unsupported tags. Now I cannot compile the markdown anymore. 

@jonasmalacofilho (how) did you managed to convert the pages?

jonasmalacofilho:
Should fix #120.

@Simn, I had no idea of what I was doing! Please review my changes and make sure I didn't screw anything up.

Also, I couldn't figure out how to implement a generic `\mbox{}` that worked with empty contents

     case [TCommand(CMbox), s = inBraces(text)]

What was my mistake?

### DIFF
--- a/convert/src/LatexCommand.hx
+++ b/convert/src/LatexCommand.hx
@@ -38,6 +38,8 @@ enum LatexCommand {
 	CTextwidth;
 	CCaption;
 	CInput;
+	CNoindent;
+	CMbox;
 
 	CTextless;
 	CTextgreater;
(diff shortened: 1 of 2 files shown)


Simn:
Looks good!

The problem with your other parsing solution is that `text` does not accept an empty string. I've made this so that it can be used conditionally because otherwise it would always accept any input. It might make sense to allow this in `inBraces` though because after the initial `{` the stream couldn't recover anyway.

Simn:
This works now, thanks for letting me know!

PR #126, opened 2015-03-31
SPOD

markknol:
I grabbed most of the info from http://old.haxe.org/manual/spod but organized it in different sections, removed the "you", created nice lists and tables etc. It's quite a big thing, I hope it's clear.

@Merelleya and @Simn  Can you check?


Simn:
Urgh and I wanted to get rid of that...

ncannasse:
Maybe that should go on the (hopefully upcoming) UFront website ? ping @jasononeil

jasononeil:
I'll have the ufront website live by WWX.  We could put it there.  But didn't we agree that these SPOD classes should go in a separate haxelib?  If so I think we should take this content (thanks @markknol!), use it as the "README.md" on the new repo, and link to it from the website.

I think many people may want to use SPOD records without using the ufront web framework, and so it would be good to keep the documentation distinct once SPOD has a place to live.

When are we going to move it out of the Standard Library?  If it's "3.3" or whatever the next major release is, maybe we can start setting up the repo and moving all the files and documentation the moment 3.2 is released.

ncannasse:
@jasononeil we decided to leave the decision of keeping/outing SPOD of std to the maintainer, which happens to be @waneck. Depending on the number of updates, we could:
- keep it in std
- move it to haxelib spod (or another name)
- move it to haxelib ufront (would ship with ufront)
I don't have any preference as long as it's properly maintained

Simn:
Actually we agreed on moving it to its own lib, it just wasn't done.

Simn:
I've generated the markdown on this branch (don't delete it!): https://github.com/HaxeFoundation/HaxeManual/blob/std-spod/md/manual/std-spod.md

This can be put wherever it's supposed to end up.

markknol:
Looks great, I hope it is getting used somewhere.

PR #133, opened 2015-04-11
Mention the range operator in the "for"-expression section

Gama11:


05-expressions.tex
```diff
@@ -220,6 +220,12 @@ \section{for}
 
 Variable \expr{v} is then available within loop body \expr{e2} and holds the value of the individual elements of collection \expr{e1}.
 
+The closest Haxe for-loops come to C-style for-loops is probably using the range operator \expr{...}. It is a binary operator taking two \expr{Int} arguments: \expr{min...max} returns an \expr{IntIterator} instance that iterates from \expr{min} (inclusive) to \expr{max} (exclusive). Note that \expr{max} may not be smaller than \expr{min}.
```
Simn: That first sentence is really out of place there. Integer iteration is only a special case of C-for-loops so I don't see why we should mention them at all (maybe they would make for a nice trivia?). Also "probably" isn't good in technical documentation.

Other than that:

* please use `\type{Int}`
* operators typically do not have arguments but operands
* expressions don't really "return" something. It is more correct to say that this syntax is typed to construct an instance of IntIterator or something.

Gama11: I mainly mentioned C-for-loops because it's the first sentence of this section ("Haxe does not support traditional for-loops known from C.").

- Ok.
- Are [Numeric operators](http://haxe.org/manual/types-numeric-operators.html) an exception then?
- It seems ok to say that an operator returns something, though? (again, that's the term used on the numeric operators page)

Simn: Ah, I didn't see the full context in the diff. Still, numeric iteration is only a special case of a C-for-loop.

Numeric operators have operands too, good catch. I'm not very happy with the "return" label in that table, but given that we use it there we can use it here as well.

Gama11: Should that be changed for the numeric operators then or should I leave it as "argument" here?

Simn: I think we should change both to operands. We should change the "return" too but I don't really know to what, so let's just leave that.

Gama11:
Addressed the things you mentioned.

Simn:
Thanks!

PR #142, opened 2015-04-16
introduction language features

markknol:
Copied from http://haxe.org/documentation/introduction/language-features.html


Simn:
Mmh, I don't really like having duplicates like that...

markknol:
Ah I thought the plan was to remove pages from the "introduction" and let them all be part in the manual. Just like the "Compiler Usage"

Simn:
Ah, yeah, I suppose it doesn't matter if we just change the link.

markknol:
Any ideas how we do that?

PR #148, opened 2015-05-23
RegExp flag `s` also applied to Flash

R32:
```haxe
class Main {	
	static function main() {
		var pat = ~/\{(.*)\}/s;
		var str = '{
			
			abc: 1,
			
			xyz: "adf",
			
			
			arr: ["a","b"]
		}
		';		
		trace(pat.match(str) ? pat.matched(1) : "false");
	}	
}
```

```bash
haxe -main Main -swf a.swf
```


Simn:
I'm surprised there would be target differences with regards to this flag. I'll merge this for now but we should check if this is actually true.

Thanks!

PR #150, opened 2015-05-27
Completion overview: mention restrictions on the usage of |

Gama11:


HaxeManual/08-compiler-features.tex
```diff
@@ -173,7 +173,7 @@ \subsection{Overview}
 \label{cr-completion-interface-notes}
 
 \begin{itemize}
-	\item The position-argument can be set to 0 if the file in question contains a pipeline \ic{|} character at the position of interest. This is very useful for demonstration and testing as it allows us to ignore the byte-counting process a real IDE would have to do. The examples in this section are making use of this feature.
+	\item The position-argument can be set to 0 if the file in question contains a pipeline \ic{|} character at the position of interest. This is very useful for demonstration and testing as it allows us to ignore the byte-counting process a real IDE would have to do. The examples in this section are making use of this feature. Note that this only works after dots and opening parentheses (\ic{.|} and \ic{(|}), since everywhere else \ic{|} is treated as a bitwise or operator.
```
Gama11: Might not be technically correct since `|` means something else in `case` expressions? What's that operator called, just "or"?

Simn: "bitwise or operator" this has to be "bitwise or-operator" or the `or` has to be emphasized in some way

Simn: I would go for "Note that this only works in places where `|` is not valid syntax otherwise, e.g. after dots or opening parentheses."

Gama11: Ok. I was thinking that might be a bit vague (raises the question of where `|` *is* valid syntax otherwise).

Gama11: Also, I don't think that's technically correct, `|` isn't valid syntax after the return type of a function for example, but position completion still doesn't work there:

```haxe
function toString():String|
{
   return "test";
}
```

`Unknown identifier : toString`

Simn: That's because we're parsing a type there and not an expression.

At this point this is going into parser internals. Given that `|` is really only meant as a debugging helper I don't think we should spend too much time thinking about it.

Gama11: Alright. Still, especially debugging features can be frustrating if they don't work as expected because of a misunderstanding. ;)

Gama11:
Updated.

Simn:
Thanks!

PR #151, opened 2015-05-27
Completion: mention the minimum required Haxe version

Gama11:


Simn:
Can you use `\since{3.2.0}` instead?

Gama11:
Done. Didn't know that was a thing. :)

Simn:
Thanks!

PR #152, opened 2015-05-27
Access Control: mention that target does not respect imports

Gama11:
Easy to trip over, see https://github.com/HaxeFoundation/haxelib/issues/214


Simn:
Thanks (again)!

PR #153, opened 2015-05-27
Update meta and defines list to Haxe 3.2.0

Gama11:
I was mostly going through the list comparing --help-metas and --help-defines with these, looking for added / removed entires and ones that changed order. That means I could easily have missed some changed descriptions.


Gama11:
@Simn Any changes required here?

markknol:
Could you also apply the new defines to the target detail pages?

Simn:
Is this just taken from common.ml?

Gama11:
@markknol Sure, didn't know that was a thing.

@Simn from the command-line - so pretty much yes, except for the Internal metadata which doesn't show in --help-metas.

Simn:
We should have some tool to generate this list from the command line output...

Gama11:
I thought the same while doing this, it's very tedious and error-prone to update this manually.

However, it's somewhat tricky because there are random links / formatting in there, like `\{tref}` and `\expr{expr}`.

markknol:
In this table, http://haxe.org/manual/cr-metadata.html there are extra references to other pages.

PR #155, opened 2015-05-28
Update the target specific metadata lists as well

Gama11:
See #153.
Also removed the "Target" column for these lists since it's redundant.


Simn:
Thanks!

markknol:
I was thinking, could we make the full table as separate asset, and put a selection of it on the detail pages? I remember there is something for code assets too `/haxe{Asset.hx#12-17}`

Like: `/table{Defines.tex#js}` (or something)

PR #159, opened 2015-06-01
Update Javascript target section

markknol:
* Added debugging section, based on old doc.
* Moved `__js__` section to new page. 
   * Also, removed the word magic, I prefer "raw injection" as seen in upcoming c# target detail section.
* Removed a line that clarified i need a script tag. That's not always the case, I think the paragraph is clear enough.

### DIFF
--- a/HaxeManual/12-target-details.tex
+++ b/HaxeManual/12-target-details.tex
@@ -110,7 +110,6 @@ The standard compilation options also provide more Haxe sources to be added to t
 
 \begin{itemize}
 	\item To add a \tref{Haxelib library}{haxelib}, use \expr{-lib <library-name>}. There are many externs for popular native libraries.
-	\item To add another class path, use \expr{-cp <directory>}.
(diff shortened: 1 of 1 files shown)


HaxeManual/12-target-details.tex
```diff
@@ -108,6 +106,16 @@ \subsection{Using external JavaScript libraries}
 extern class Video {
 ..
 \end{lstlisting}
+The standard compilation options also provide more Haxe sources to be added to the project:
+
+\begin{itemize}
+	\item To add a \tref{Haxelib library}{haxelib}, use \expr{-lib <library-name>}. There are many externs for popular native libraries.
+	\item To add another class path, use \expr{-cp <directory>}.
```
Simn: I'm fine with mentioning -lib because it is related to JS externs, but -cp seems a bit out of place here.

Simn:
Thanks!

PR #163, opened 2015-06-02
XML

markknol:
Another copy/paste from old.haxe. 

Contains:
* Getting started with Xml
* Encoding Xml
* Parsing Xml
* Fast Xml access

I provided encoding and parsing also as separate pages to keep in line with Json, while parsing is already covered in getting started.


HaxeManual/10-std.tex
```diff
@@ -597,6 +597,170 @@ \subsection{Serialization format}
...
+
+An root node element can also be created by parsing a \type{String} containing the XML data.
+\begin{lstlisting}
+var root = Xml.parse('<root />').firstElement();
+trace(root); // <root />
+\end{lstlisting}
+
+\paragraph{Creating child elements}
+
+Adding child elements to the root can be done using the \expr{createElement} method.
```
Simn: You mean `addChild`.

HaxeManual/10-std.tex
```diff
@@ -597,6 +597,170 @@ \subsection{Serialization format}
...
+trace(xml.firstChild().nodeValue); // Haxe is great!
+\end{lstlisting}
+
+The difference between \expr{firstChild} and \expr{firstElement} is that the second function will return the first child with the type \type{Xml.Element}. 
+
+\paragraph{Iterate on Xml elements}
+We can as well use other methods to iterate either over children or elements.
+\begin{lstlisting}
+for (child in xml) {
+	// iterate on all childs.
```
Simn: children

HaxeManual/10-std.tex
```diff
@@ -597,6 +597,170 @@ \subsection{Serialization format}
...
+}
+for (att in xml.attributes()) {
+	// iterator on all attributes.
+}
+\end{lstlisting}
+
+\subsection{Parsing Xml}
+\label{std-Xml-parsing}
+
+Use the \expr{Xml.parse} static method to parse \emph{XML} data and obtain a Haxe value from it:
```
Simn: This is a change of tone because you're addressing the reader directly here. Suggestion: "The static method Xml.parse can be used to ..."

This also applies to some content below.

HaxeManual/10-std.tex
```diff
@@ -597,6 +597,170 @@ \subsection{Serialization format}
...
+
+\begin{lstlisting}
+var xml = Xml.parse('<root>Haxe is great!</root>').firstElement();
+trace(xml.firstChild().nodeValue);
+\end{lstlisting}
+
+\subsection{Encoding Xml}
+\label{std-Xml-encoding}
+
+Use the \expr{xml.toString()} method to encode a Haxe value into a \emph{XML} string:
```
Simn: This is a bit convoluted, let's instead talk about "obtaining the String representation" or something.

HaxeManual/10-std.tex
```diff
@@ -597,6 +597,170 @@ \subsection{Serialization format}
...
+\begin{lstlisting}
+var xml = Xml.createElement('root');
+xml.addChild(Xml.createElement('child1'));
+xml.addChild(Xml.createElement('child2'));
+
+trace(xml.toString()); // <root><child1/><child2/></root>
+\end{lstlisting}
+
+\subsection{Fast Xml access}
```
Simn: I'm fine with documenting it but we can _not_ call this "fast". I'm fine with "simplified" or something to that extent as long as it doesn't suggest that this kind of access is in any way run-time-fast (it's not).

HaxeManual/10-std.tex
```diff
@@ -597,6 +597,170 @@ \subsection{Serialization format}
...
+\label{std-Xml-Fast}
+
+While the base \type{Xml} API is easy to use, it is sometimes a bit painful to do all the checks of \type{Xml} correctness and can result in verbose code. The \href{http://api.haxe.org/haxe/xml/Fast.html}{haxe.xml.Fast} API is here to help by providing a fast dot-syntax access to the most common \type{Xml} methods.
```
Simn: Can we use something other than "painful"? That's not something that should appear in a technical documentation in my opinion.

Maybe just lose the first sentence entirely?

Also "The API is here" sounds a bit silly. Would be better to go for "The API exists to ..."

HaxeManual/10-std.tex
```diff
@@ -597,6 +597,170 @@ \subsection{Serialization format}
...
+	\item[\expr{.has.<name>}] check the existence of an attribute.
+	\item[\expr{.elements}] the list of all sub-elements (which are the nodes with type \type{Xml.Element}).
+	\item[\expr{.node.<name>}] access to the first sub element with the given name. An exception is thrown if the element doesn't exists.
+	\item[\expr{.nodes.<name>}] returns a List of elements with the given name.
+	\item[\expr{.hasNode.<name>}] check the existence of a sub node with the given name.
+	\item[\expr{.innerData}] returns the inner \ic{PCDATA} or \ic{CDATA} of the node. An exception is thrown if there is no data or if there not only data but also other nodes.
+	\item[\expr{.innerHTML}] returns the XML string built with all the sub nodes, excluding the current one.
+\end{description}
+
+\paragraph{Self-closed nodes}
```
Simn: This entire section confuses me, I'm not quite sure what it is about and why this would be a special case.

markknol: Isn't this how the syntax of Fast works?

Simn:
Could we split the xml.Fast part off for now and just merge the rest of the documentation? I really want to talk about this thing before documenting it.

HaxeManual/10-std.tex
Simn: I'm talking about the "Self-closed nodes" part.

Simn:
Thanks a lot!

Issue #165, opened 2015-06-06
Interfaces inheritance


RealyUniqueName:
There is no description in `Interfaces` section how intefraces can extend each other.

PR #166, opened 2015-06-07
Interfaces can be extended

markknol:
Closes https://github.com/HaxeFoundation/HaxeManual/issues/165

### DIFF
--- a/HaxeManual/02-types.tex
+++ b/HaxeManual/02-types.tex
@@ -332,9 +332,9 @@ Interface fields are not limited to methods. They can be variables and propertie
 
 \haxe{assets/InterfaceWithVariables.hx}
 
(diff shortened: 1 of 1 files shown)


Simn:
Can you also add an example how to extend multiple interfaces? Also from -> by

markknol:
Is that : extends A extends B
Didnt know that was possible?

Simn:
See, that's why we document things. :)

We could also add a trivia that the syntax used to use `implements I` but was changed for Haxe 3. It's somewhat relevant for macro people because internally the extended interfaces are still stored in `cl_implements`, not `cl_extends`. But I can add that trivia later if you prefer.

markknol:
Ha! And thats why I like contributing; to discover such unknown features. 

Please add this trivia, thats nice information :+1:

PR #167, opened 2015-06-10
Compile the HaxeManual code examples on Travis

Gama11:
I had to make some changes to make some of the examples testable, like renaming the classes to `Main` (this also makes things more consistent) and moving macros into separate modules.

All targets on both Haxe 3.2.0 and the dev branch are tested - they're [all green for the dev branch right now](https://travis-ci.org/Gama11/HaxeManual/builds/66239660), but not 3.2.0 due to several issues. Not sure what the best way to deal with that is:

- add the examples in question to the list of excluded examples
- allow failures for all 3.2.0 builds in the travis file
- wait for a Haxe 3.2.1 release? :)

The first argument to the neko test runner specifies the target. All subsequent arguments specify a specific example file that should be built. If any are specified, only those will be built (instead of building all when none are specified).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,18 @@ haxe:
   
 env: 
   - TARGET=swf
+  - TARGET=as3
   - TARGET=js
   - TARGET=neko
   - TARGET=cpp
+  - TARGET=cs
+  - TARGET=java
+  - TARGET=python
 
 install:
   - haxelib install hxcpp > /dev/null
+  - haxelib install hxcs > /dev/null
+  - haxelib install hxjava > /dev/null
   - haxelib install random > /dev/null
   - haxelib list
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: haxe
 
 haxe:
   - "3.2.0"
-  - "3.1.3"
   - development
   
 env: 
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ env:
   - TARGET=python
 
 install:
+  - sudo apt-get install mono-devel mono-mcs -qq
   - haxelib install hxcpp > /dev/null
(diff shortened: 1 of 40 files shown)


Simn:
Holy cow, thanks a lot!

We'll have to check references to the class names in the text because now everything is called Main. I'm fine with doing that later though as long as we don't forget about it.

Gama11:
Any thoughts on the 3.2.0 failures? The builds aren't too useful if they're always red.

Simn:
Oh, I didn't even check what fails because I was so happy... :)

.travis.yml
```diff
@@ -0,0 +1,28 @@
...
+  - TARGET=as3
+  - TARGET=js
+  - TARGET=neko
+  - TARGET=cpp
+  - TARGET=cs
+  - TARGET=java
+  - TARGET=python
+
+install:
+  - sudo apt-get install mono-devel mono-mcs -qq
```
andyli: @Gama11 Let's use the [apt addon](http://docs.travis-ci.com/user/apt/) instead. It will allow the builds to start much faster with the container-based infrastructure.
e.g.
```yml
addons:
  apt:
    packages:
      - mono-devel
      - mono-mcs
```

Gama11: Thanks, didn't know you could do that. :)

markknol:
This is really nice, great stuff @Gama11 ! :+1: :+1:

markknol:
Hey, now you're on it, how hard would it be to have the try.haxe examples also unit tested (for js/flash target)

Gama11:
@markknol Sure, why not, I might look into that.. Should be a lot simpler than the test runner for this, with all the special cases.

markknol:
Btw, does this also unit tests the actual manual converter or only the examples?

Gama11:
Just the examples at the moment.

PR #168, opened 2015-06-10
Types: interface should be lowercase, Debugable -> Debuggable

Gama11:


Simn:
Waiting to see if travis does something. ;)

markknol:
Ah this is my bad, sorry :cry:

PR #169, opened 2015-06-10
Use apt addons and exclude failing examples

Gama11:


Simn:
Ah sorry, I excluded them too. I'll add the other change manually. Thanks!

Gama11:
No problem. Your solution is better anyway. :)

PR #174, opened 2015-06-11
Travis: less excluded examples

Gama11:
- moved some examples from `excludedExamples` to `requiredFailures` where they should be
- added some hackery to insert snippets that aren't compiable on their own into a .hx file as modules / the body of the main function (:D)
- rewrote `Color.hx` to be compilable


Simn:
Thanks!

Issue #146, opened 2015-05-19
Document @:optional


markknol:
From forum https://groups.google.com/forum/#!topic/haxelang/MXe0BbALOAw
It might be a good idea to place a bit info of `@:optional` in [this](http://haxe.org/manual/types-nullability-optional-arguments.html) and [this](http://haxe.org/manual/types-anonymous-structure.html) or [this](http://haxe.org/manual/type-system-typedef.html) page.

PR #175, opened 2015-06-14
@:optional for Typedef

markknol:
* added "optional fields" paragraph
* 2 spaces indent
* fix indent of type-system-unification

closes https://github.com/HaxeFoundation/HaxeManual/issues/146


Simn:
That's a good start, but it's going to require some more explanation. I'll merge for now but let's keep the issue open.

markknol:
Yes, to be honest I was stuck at "you can mark stuff optional with @:optional"

PR #176, opened 2015-06-14
Add more links to API documentation

### LINKED ISSUE #173: Add links to API docs
I think we need crosslinks between the manual and the API docs. Both need links to the other one.
Especially on the standard library section.

markknol:
Closes #173 

I tried to make a line on each end of section to the API docs and also converted some \type to \href in the text.

Also removed line of text on first section, otherwise NoContent state does not seem to work. Now it nicely makes a list of everything.

### DIFF
--- a/HaxeManual/10-std.tex
+++ b/HaxeManual/10-std.tex
@@ -2,8 +2,6 @@
 \label{std}
 \state{NoContent}
 
-Standard library
-
 \section{String}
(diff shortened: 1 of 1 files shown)


Simn:
The original idea of \type was to automatically insert API links so this wouldn't be necessary. The problem is that we also use it for types from the examples which obviously don't have an API page.

PR #178, opened 2015-06-15
Update HelloWorld.hx

sneurlax:
Documentation in 01-introduction.tex says that this file should be saved as HelloWorld.hx -- but currently you need to use Main.hx

This change reflects that


markknol:
I think the documentation should say it needs to be Main.hx

Simn:
This has been fixed, thanks!

PR #183, opened 2015-07-27
add missed defines

fzzr-:


fzzr-:
And i miss some more
`HXCPP_NO_M32`
`EMCC_CORES`

PR #190, opened 2015-08-17
Updated "Using external JavaScript libraries"

andyli:
Updated "Using external JavaScript libraries" to include info on the new jQuery/swfobject externs.

### DIFF
--- a/HaxeManual/12-target-details.tex
+++ b/HaxeManual/12-target-details.tex
@@ -69,33 +69,54 @@ To display the output in a browser, create an HTML-document called \ic{index.htm
 \subsection{Using external JavaScript libraries}
 \label{target-javascript-external-libraries}
 
-The Haxe Standard Library comes with many externs for the JavaScript target. They provide access to the native APIs in a type-safe manner.
-The \tref{externs mechanism}{lf-externs} assumes that the defined types exist at run-time but assumes nothing about how and where those types are defined. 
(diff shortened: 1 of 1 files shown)


markknol:
Looks good to me! 

Offtopic question; what does `@:selfCall` do?

andyli:
It denotes that function is a direct call to the object. In case of applying `@:selfCall` to the constructor, it avoids generating `new`. e.g. `new JQuery("div")` will be outputted as `$("div")` instead of `new $("div")`. Maybe I should add that info to the manual too.

nadako:
@markknol see https://github.com/HaxeFoundation/haxe/issues/3441

markknol:
Nice! Would be good idea to document. Is this JS specific?

andyli:
Just added some doc on `@:selfCall`. Yes, the current implementation of it is JS specific.

Simn:
Did you check if the markdown converter handles the table correctly? That was a bit finicky so I might have to adjust some things.

andyli:
I've just checked the output and fixed a few things.
The markdown output is fine now, but the ebook ones do not. The [table in introduction](https://github.com/HaxeFoundation/HaxeManual/blob/master/HaxeManual/01-introduction.tex#L16-L30) is not correctly handled in the .epub either. Is that expected?

Simn:
I couldn't figure out what's causing that yet. IIRC some tables were working and other were not. It's not very important though, so if the markdown output works I'm fine with merging this. Thanks!

PR #192, opened 2015-08-28
Add instantiation reification syntax

markknol:
I had [this question ](http://stackoverflow.com/questions/32179173/how-to-declare-instantiation-in-haxe-macro-function) about instantiation within a macro function. It seems to exist but it's not very documented well. I found the [Expression Reification page](http://haxe.org/manual/macro-reification-expression.html) only says `function $name() { }` from which you _could guess_ the syntax exist.

I'm not sure if we should also mention that a full path should be given here.

### DIFF
--- a/HaxeManual/09-macros.tex
+++ b/HaxeManual/09-macros.tex
@@ -105,6 +105,7 @@ This kind of reification only works in places where the internal structure expec
 	\item field name \expr{\{ \$name: 1\} }
 	\item function name \expr{function \$name() \{ \}}
 	\item catch variable name \expr{try e() catch(\$name:Dynamic) \{\}}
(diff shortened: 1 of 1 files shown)


Simn:
Does that really work? I thought `new` required a TypePath: http://api.haxe.org/haxe/macro/TypePath.html

markknol:
Ah sorry yes that is true, maybe `instantiation \expr{new \$typePath() \{ \}}` is more clear then?

Simn:
It's still in the wrong section because that listing deals with places where the internal structure expect strings.

markknol:
Oh, I thought that list mentions the syntax since Haxe 3.1.0. 
What would be the correct place to put it?

Simn:
A new paragraph I suppose.

PR #196, opened 2015-09-06
Fixed " conversion of single quoted empty string

MrMormon:
Changed \expr{''} to \expr{'{'}} in Constants section

See http://haxe.org/manual/expression-constants.html right before 'bar' for the problem. This is undesired behavior in this particular instance in the manual.


Simn:
Thanks!

MrMormon:
You're welcome! The next build of the manual should propagate to markdown, html etc. right?

Simn:
It doesn't build automatically at the moment.

MrMormon:
Should I propagate it myself or is that all part of the website's regular maintenance?

Simn:
We'll take care of it.

PR #201, opened 2015-09-29
Fix templating example

kevinresol:


markknol:
For contributions please edit the .tex file (https://github.com/HaxeFoundation/HaxeManual/tree/master/HaxeManual). The .md files are generated from it.

PR #202, opened 2015-09-29
add missing script close tag

chromy:
Unless the the script tag is properly closed the example seems not to work under Chrome v48


Simn:
Thanks!

Issue #54, opened 2014-09-02
Type inference - open/closed/const structures


Simn:
Someone has to do it eventually.

Simn:
There might be changes here (#3192) so I'll refrain from documenting this for now.

PR #208, opened 2015-10-04
Added Inference status section

markknol:
I gathered some information from several issue pages, please double check.

Closes https://github.com/HaxeFoundation/HaxeManual/issues/54


Simn:
I'm still not sure if this behavior should be surfaced to the user. It's pretty much an implementation detail with some consequences.

markknol:
Are you saying that In that case it shouldnt be catergorized in statusses like this?

Simn:
Const status has been removed so we have to update this.

markknol:
Can you take care of it? 
Also because you weren't sure if his was relevant enough to communicate.

Simn:
Okay, let's not document this for now, I really think it's an implementation detail.

Simn:
Keep the branch though.

Issue #211, opened 2015-10-12
Python target also requires file name.


shohei909:
```
Depending on the nature of the target, the argument value is either a file name (for -js, -swf and neko) or a directory path.
```
https://github.com/HaxeFoundation/HaxeManual/blob/574e150a95668dabdfeae3fd062595a6b9cb03a9/output/HaxeManual/website/compiler-usage.md

PR #212, opened 2015-10-12
fix for "Python target also requires file name"(#211)

shohei909:
fix for https://github.com/HaxeFoundation/HaxeManual/issues/211


Simn:
Thanks!

PR #219, opened 2015-11-28
JavaScript Lib.alert was deprecated.

ypid:
Main.hx:8: characters 12-21 : Warning : Lib.alert() is deprecated, use Browser.alert() instead


Simn:
Thanks!

PR #233, opened 2016-02-17
Change the "Based on work" link

JefePapaya:
The source as is displays the single link on the browser (tested on Firefox and Chrome)  "\[https://github.com/HaxeFoundation/HaxeManual](https://github.com/HaxeFoundation/HaxeManual)". So, as expected, this redirects the user to a 404 page on github. I don't think this is intendend.

I changed the link description to "Haxe Manual on GitHub", but it's just a suggestion. You may have a better description for it.

### DIFF
--- a/HaxeManual/01-introduction.tex
+++ b/HaxeManual/01-introduction.tex
@@ -84,7 +84,7 @@ Most of this document's content was written by Simon Krajewski while working for
 
 The Haxe Manual by \href{http://haxe.org/foundation}{Haxe Foundation} is licensed under a \href{http://creativecommons.org/licenses/by/4.0/}{Creative Commons Attribution 4.0 International License}.
 
-Based on a work at \href{https://github.com/HaxeFoundation/HaxeManual}{https://github.com/HaxeFoundation/HaxeManual}.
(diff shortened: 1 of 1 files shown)


Simn:
I don't understand what this is supposed to address. Where does a 404 happen?

JefePapaya:
On the page http://haxe.org/manual/introduction-license.html when you click the link "Based on a work at <LINK>"

You should try this on your browser.

Simn:
Hmm, the generated markdown is fine: https://github.com/HaxeFoundation/HaxeManual/blob/master/output/HaxeManual/website/introduction-license.md

This might be an issue with haxe-markdown @dpeek.

JefePapaya:
Were you able to see the issue in the page?
Probably happens when you give another link as the target link's display name

Simn:
Yes but I don't think it should happen, it's likely a problem with the haxe-markdown conversion and your change merely sidesteps the issue. I prefer the current version so I'll go ahead and close this.

JefePapaya:
Yep, that makes sense. Hope this helps fixing the issue.

Simn:
Yes, thanks for letting me know about the 404!

PR #239, opened 2016-03-05
Built-in Compiler Metadata: fix misleading abstract link

Gama11:
@:abstract is for abstract classes and has nothing to do with Haxe abstracts.


ibilon:
Thanks!

PR #241, opened 2016-03-07
Update 08-compiler-features.tex

tecteun:
@:functionCode is also supported on cs target

@see:
https://github.com/HaxeFoundation/haxe/blob/development/gencs.ml#L2172


Simn:
Thanks!

PR #242, opened 2016-03-08
[TraivsCI] Install hxcpp from git when testing Haxe dev

Gama11:
Should fix the Travis cpp/dev failures.

Let the Travis build finish before merging to see if this actually works.


Gama11:
Hm, still failing..

PR #252, opened 2016-04-06
Static analyzer modules: highlight names

Gama11:
I hope this is valid Latex. :)


Simn:
Can you remove the `unreachable_code` one while you're there?

Gama11:
Done. What happened to that?

Simn:
I've changed some internal handling so it no longer worked. Also there are quite a few problems with intentional unreachable code that comes from conditional compilation. I might revisit it again soon because I still think it can be useful.

PR #253, opened 2016-04-10
$a{} type seemed to be  Array<Expr> -> Array<Expr> or Array<Expr> -> Expr

shohei909:
I got `haxe.macro.Expr should be Array<haxe.macro.Expr>` error when I use `$a{expr}`.


Simn:
Yes that's correct, thanks!

PR #255, opened 2016-04-21

markknol:
Create maps with array comprehension

### DIFF
--- a/HaxeManual/06-language-features.tex
+++ b/HaxeManual/06-language-features.tex
@@ -528,6 +528,19 @@ The loop expression can be anything, including conditions and nested loops, so t
 
 \haxe{assets/AdvArrayComprehension.hx}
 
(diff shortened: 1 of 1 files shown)


Gama11:
"Using array comprehension to create maps" sounds quite strange to me. Shouldn't that be called _map_ comprehension in that case?

Simn:
Yeah this is not really a good place for it.

markknol:
Is this any good? Otherwise lets close it.

Simn:
The content is fine but it's strange to have this under array comprehension.

ibilon:
How about making a page "Map comprehension".

Simn:
I'll go ahead and close this because it no longer merges.

markknol:
Good, because we already have it here too: https://haxe.org/manual/lf-map-comprehension.html

PR #259, opened 2016-06-13
Trace

markknol:


markknol:
I've updated the examples so they actual compile (except on cpp but I have no idea this fails on everything)

Simn:
I'll go ahead and merge this though I still think the PosInfos part should be moved elsewhere.

markknol:
Thanks! 
Good point about the PosInfos, do you mean to a separate section?

PR #270, opened 2016-12-29
Move documentation for type extensions to the Anonymous Structures se…

### LINKED ISSUE #269: Document anonymous structure inheritance/type spread operator
I saw this syntax in [`Type.hx`](https://github.com/HaxeFoundation/haxe/blob/de17354927f84d46320b57bf69e3d469e2352211/std/haxe/macro/Type.hx#L396) and had been wondering how to do the same thing in my own code:

```haxe
typedef ClassType = {> BaseType,
```

I don’t see anything glaringly obvious about this in [Anonymous Structure](https://haxe.org/manual/types-anonymous-structure.html). Is this something I’m allowed to do in my own code and what are the rules for it? If it really is as self explanatory as it appears, I could try writing a snippet about it. But I don’t actually know what to call it ;-).

binki:
…ction.

Fixes #269.

### DIFF
--- a/HaxeManual/02-types.tex
+++ b/HaxeManual/02-types.tex
@@ -482,6 +482,12 @@ class Path {
 }
 \end{lstlisting}
 
+You may also use \tref{Extensions}{types-structure-extensions} to ``inherit'' fields from other structures.
+
+\begin{lstlisting}
+typedef Point3 = { > Point, z : Int }
+\end{lstlisting}
+
 
 \subsection{JSON for Structure Values}
 \label{types-structure-json}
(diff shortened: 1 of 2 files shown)


Simn:
Thanks, that looks good! My only concern is that we're removing an existing page, which is going to lead to potential 404 links. @ibilon: Can we deal with that somehow?

ibilon:
If a simple redirection is enough then it's just a matter of adding the old/new link to https://github.com/HaxeFoundation/haxe.org/blob/staging/src/generators/Redirections.hx#L10

Simn:
Ah, perfect.

binki:
@ibilon  Hmm, though it’d be nice to be able to write the manual-specific redirections in this repo instead of needing cross-repo coordination when merging.

Simn:
Well, that would be technically better, but I don't think it's worth any effort given that we move stuff like once a year. In fact, I think this is the first time we remove a page.

PR #279, opened 2017-03-22
document variable shadowing

nadako:


Simn:
Looks good. I wonder if we should mention this case while we're at it:

```haxe
var a = 1;
function f() {
    trace(a);
}
var a = 2;
f(); // traces 1
```

This makes it clear that the shadowing strictly follows syntax and that you don't have to worry about re-declaring a variable that is captured by a closure.

nadako:
That's a nice example! Do you want me to add it or are you going to do it?

nadako:
BTW, maybe it should have a separate sub-heading so we can link to these explanations.

Simn:
Feel free to add it to this PR.

And I agree, we should have more paragraphs. You can just add `\paragraph{Variable Shadowing}` for that.

nadako:
ok, done

nadako:
thanks! publish soon?

Simn:
Uhm, I thought this published automatically...

ibilon:
Not anymore, the haxelib repository needs to update the haxemanual submodule (which will trigger travis).
I'll do that right now.

nadako:
That's nice, actually, thanks!

PR #282, opened 2017-04-19
Add pointer to class methods from function call docs

techtonik:


markknol:
Hi, thanks for your contribution!

It is better to use `\tref{Class Methods}{class-field-method}` or `\Fullref{class-field-method}` for internal links, can you please change that?

techtonik:
Done.

PR #283, opened 2017-05-01

markknol:
Some bits on haxelibs and how to use flags from code


Simn:
Can we stick to "defines" instead of "directives"? I think that's what we use everywhere else.

markknol:
Yes, done.

PR #284, opened 2017-05-02
@:functionCode and @:functionTailCode descripton

dkolyesnik:
Added @:functionCode and @:functionTailCode description


nadako:
Isn't that list autogenerated from the compiler description? @markknol ?

profelis:
```
$ haxe --help-metas | grep "@:function"                                                       
 @:functionCode         : Used to inject platform-native code into a function
 @:functionTailCode     :  (cpp only)
```
@nadako

markknol:
No I copy/pasted it once from there but added some markup/links to it. So this is a good addition. 

We might want to double check if the rest is still up to date.

markknol:
Thanks!

If any other metadata needs to be updated, please open a new PR.

PR #287, opened 2017-05-24
Alternative typedef declaration

RblSb:
from http://old.haxe.org/ref/type_advanced


Gama11:
Might be worth mentioning the downside of short notation (fields can't have doc comments).

RblSb:
@Gama11 I checked, it seems everything works. Can you give an example?

ibilon:
Comments aren't used as documentation in the short notation.
Also you can't use metadata.

PR #290, opened 2017-06-06
Fix client side code

fantostisch:
The original code gives a compile error:
Client.hx:4: characters 61-62 : Unexpected }
Client.hx:4: characters 63-64 : Unexpected )


ibilon:
Sorry for the delay, and thanks for the contribution.

Issue #288, opened 2017-05-24
Use Travis to commit generated markdown files


markknol:
We now manually convert and push to https://github.com/HaxeFoundation/HaxeManual/tree/master/output/HaxeManual/website

Would be nice if that step could be automated.

[bonus points if you find out why travis takes ~40 minutes to build in Travis]

markknol:
Seems like travis is slow because of some latex related plugin.
```
The command "sudo apt-get install texlive-xetex texlive-latex-extra -y" failed and exited with 100 during .
```

PR #291, opened 2017-06-06
Don't download/generate PDF in Travis anymore

markknol:
* Latex libraries really slow down Travis. The resulting pdf-file isn't used anywhere, so that doesn't make it worth the waiting time (which is ~1.5 hours).
 * For convenience I created two hxml files.
 * Travis now only generates the markdown pure for testing and runs each example code in assets-folder on all targets.
 * Temporary fixes https://github.com/HaxeFoundation/HaxeManual/issues/288
* Unrelated, but let's not use git haxelib versions if it is not needed

This build should take less time: https://travis-ci.org/HaxeFoundation/HaxeManual/builds/240126143 Otherwise this PR is pointless.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,10 @@ sudo: false
 matrix:
   include:
     - sudo: required
-      install:
-        - haxelib install convert-markdown.hxml
+      install:		
+        - haxelib git hxargs https://github.com/Simn/hxargs.git		
+        - haxelib git hxparse https://github.com/Simn/hxparse.git
         # spellcheck
-
         - nvm install 6.4
         - npm install -g markdown-spellcheck
       script:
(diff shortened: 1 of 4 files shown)


andyli:
Unless it is decided that we will never generate the pdf again, I would insist keeping the pdf generation in CI. Or else we will break it someday and will be super hard to get it working again.

Simn:
I would prefer keeping it as well. The tex compilation might catch some problems that aren't picked up by the markdown conversion.

Maybe there's a different way to speed up the CI.

markknol:
Ok In that case I'll close it. I would love to see it being speed up somehow.

Gama11:
Couldn't it just be a separate job, so it doesn't matter that it's slow?

Issue #34, opened 2014-05-30
Function bindings and optional arguments


frabbit:
that special case needs some docs (see https://github.com/HaxeFoundation/haxe/issues/3082)

PR #293, opened 2017-06-15
Binding functions with optional arguments

RealyUniqueName:
Second attempt to address HaxeFoundation/haxe#6351 and #34
But i'm bad with english, so feel free to propose better text )


Simn:
Note: If you want the example to fail, you have to add it to the list here: https://github.com/HaxeFoundation/HaxeManual/blob/master/tests/RunTravis.hx#L57

RealyUniqueName:
Done!

RealyUniqueName:
> the original function becomes a non-optional argument of the result function

Actually it becomes optional argument of the result function: https://try.haxe.org/#847b5

PR #295, opened 2017-07-06
Update GenericTypeParameter.hx

romamik:
The old code is not compiling since haxe 3.3. It is a little bit confusing.


ibilon:
Thanks.

Simn:
Why was this not failing on our CI?

ibilon:
It's actually only a warning on haxe 3.3+

PR #298, opened 2017-07-26
Start on some hxcpp docs

hughsando:


Simn:
Nice!

@markknol: Do you think we should remove mdspell? I think you complained about it before, and it really seems to be more of a hassle than helpful...

hughsando:
There also seems to be a bunch of generated files, which I have not checked in.
Not sure of the procedure there.

markknol:
This is great @hughsando !

@Simn @andyli Can we also change it to warnings instead of errors? Spell check is nice but now cost more time to maintaining it than it actually helps. Also should be task for author and reviewers.

Simn:
I've merged this locally to try and deal with the .mdspell problems.

PR #316, opened 2018-01-05
Fix discrepancy between text and asset Haxe code

vincentb1:
In a nutshell, `array` in the text body is renamed to `myArray` in order to be aligned with the sample code.


markknol:
Nicely spotted!

PR #325, opened 2018-02-20
Added xml.Access documentation

markknol:
We still have this documentation. Since Fast survived in Haxe 4, maybe it should be in docs?

If so, we need to resolve conflicts of it.


Simn:
I guess...

Simn:
The content here looks good. We renamed it to `haxe.xml.Access` for Haxe 4. Could you update so we can merge this?

markknol:
I've updated the copy. Also added a note that this was known as Fast, but now we made it Slow.

Simn:
Could you update the branch to see if the CI turns green?

HaxeManual/10-std.tex
```diff
@@ -749,21 +749,96 @@ \subsection{Encoding Xml}
...
+Please note that we cannot access self-closed nodes as "regular" ones. 
+For example, provided this XML:
+
+\begin{lstlisting}
+var xml = Xml.parse("
+  <myNode1></myNode1>
+  <myNode2/>
+");
+
+var access = new haxe.xml.Access(xml.firstElement());
```
Simn: This example is borked because `xml.FirstElement()` would refer to the `myNode1` here. Might want to wrap the content that's parsed above in `<xml> ... </xml>` for this to make sense.

HaxeManual/10-std.tex
```diff
@@ -749,21 +749,96 @@ \subsection{Encoding Xml}
...
+
+var myNode1Value = access.node.myNode1.innerData; 
+// no problem
+
+var myNode2Value = access.node.myNode2.innerData; 
+// ERROR, an exception is thrown!
+\end{lstlisting}
+
+In case the application might get self-closed nodes, check \expr{hasNode.innerData} before and default to an alternative value.
```
Simn: "application might get" sounds really strange. I wouldn't talk about applications at all here.

"If the XML might contain self-closed nodes"

Also I wonder if we should just return `null` in this case instead of throwing an exception. Seems a bit awkward API-wise...

HaxeManual/10-std.tex
markknol: It does work here: https://try.haxe.org/#330f2

Simn: There it refers to the `<user>` node. Here it refers to the `<myNode1>` node and then you try to access `node.myNode1` which doesn't make sense.

PR #333, opened 2018-04-10
Static analyzer -> Static Analyzer

Gama11:
All the other pages here use title-case, looks a bit strange with the lowercase `a`:


Gama11:
Actually, there seem to be more cases of this..

Gama11:
#334

PR #338, opened 2018-05-17
Add missing metas

T1mL3arn:


markknol:
Nice. I think maybe we should add another column to define in which Haxe version all metas are usable.

T1mL3arn:
Two column maybe? For last Haxe version, in case if there are deprecated metas.

Simn:
The metadata section needs a review. This doesn't merge anymore and is probably not accurate anyway, so I'll go ahead and close it.

PR #341, opened 2018-06-21
Fix of the doc for Map Comprehension

Epono:
The asset `MapComprehension.hx` was just a copy of `ArrayComprehension.hx`.
There was also an error in the equivalent code for the `for` loop example (`Loop variable cannot be modified`)


markknol:
Thanks

PR #347, opened 2018-08-17
escape ampersand

romamik:
Current version is
> The output is HTML-escaped so that &, < and > become &, < and > respectively.

but it should read like this
> The output is HTML-escaped so that &, < and > become &amp;amp;, &amp;lt; and &amp;gt; respectively.


Simn:
Thanks!

PR #354, opened 2018-09-11
Document `final`

markknol:


HaxeManual/05-expressions.tex
```diff
@@ -272,6 +273,35 @@ \section{var}
...
+\label{expression-final}
+
+\since{4.0}
+
+The \expr{final} keyword allows being used in two places:
+
+\begin{itemize}
+	\item As modifier to a method, like in Java.
+	\item As replacement for \expr{var} when declaring fields.
+\end{itemize}
```
Simn: This isn't quite complete. We can use `final` in the following situations:

1. `final class` to make it non-extensible
2. `final function` to make a function non-overridable.
3. `static final x` that require immediate initialization and cannot be written to.
4. `final x` that require initialization immediately or in the constructor, and cannot be written to otherwise (note however that it _can_ be assigned to multiple times in the constructor).
5. `final` at expression-level, which is like `var` but cannot be assigned to after the initial initialization (which currently isn't required... might want to fix that still).
6. `final field` on structures. This guarantees read-only. The structure only unifies with classes or other structures where `field` is also `final`.

Gama11: Also `final interface`.

Aurel300:
This has been merged as part of #383. I cherry-picked @haxiomic's commit removing `-D old_browser`.

PR #361, opened 2018-10-28
Fix upcase after colons

archiloque:


Simn:
This isn't necessary, see https://www.grammarly.com/blog/capitalization-after-colons/.

I "think" we don't capitalize in these cases, but I'll check again to make sure we're consistent.

PR #365, opened 2018-12-02
Changed description of function

CoruNethron:
"a compound type of several arguments and one return" may be misunderstood as "one return statement", but in fact it means "one return value". Also, there might be no return value, i.e. [Any]->Void

### DIFF
--- a/HaxeManual/02-types.tex
+++ b/HaxeManual/02-types.tex
@@ -24,7 +24,7 @@ The Haxe type system knows seven type groups:
  \item[\emph{Class instance}:] an object of a given class or interface
  \item[\emph{Enum instance}:] a value of a Haxe enumeration
  \item[\emph{Structure}:] an anonymous structure, i.e. a collection of named fields
- \item[\emph{Function}:] a compound type of several arguments and one return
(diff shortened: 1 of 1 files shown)


Aurel300:
@simn Any thoughts on this? What about:

> a function which expects zero or more arguments of specific types that will return a value

I think it's fine to consider `Void` a value here so we don't complicate the short description.

player-03:
> This leads to the old "is Void a type" discussion.

Whether or not it's a _type_, I think we can all agree it's a _return type_. Which suggests a simple change:

> * **Function:** a compound type of several arguments and one return type

Or for consistency:

> * **Function:** a compound type of several argument types and one return type

Or if you're fine leaving out the reference to "compound type":

> * **Function:** a collection of argument types plus one return type

Or to make it clearer that we're talking about the function type and not a function definition:

> * **Function:** a function signature, consisting of a number of argument types plus a return type

...but that doesn't quite fit the rest of the page. The rest of the page draws less distinction between the type and the value stored in a variable of that type. Perhaps this would fit better:

> * **Function:** a Haxe function, with known argument types and return type

PR #368, opened 2018-12-03
Update 02-types.tex

paul59:
Grammatical changes and some clarification

### DIFF
--- a/HaxeManual/02-types.tex
+++ b/HaxeManual/02-types.tex
@@ -1,7 +1,7 @@
 \chapter{Types}
 \label{types}
 
-The Haxe Compiler employs a rich type system which helps detecting type-related errors in a program at compile-time. A type error is an invalid operation on a given type such as dividing by a String, trying to access a field of an Integer or calling a function with not enough (or too many) arguments.
+The Haxe Compiler employs a rich type system which helps to detect type-related errors in a program at compile-time. A type error occurs when an invalid operation is performed on a given type. For example, attempting to divide by a String, trying to access a field on an Integer or calling a function with the wrong number of arguments are all type errors.
 
 In some languages this additional safety comes at a price because programmers are forced to explicitly assign types to syntactic constructs:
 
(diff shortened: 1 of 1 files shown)


HaxeManual/02-types.tex
```diff
@@ -46,27 +46,27 @@ \section{Basic Types}
...
+Basic types are not \tref{classes}{types-class-instance} in Haxe, instead they are implemented as \tref{abstract types}{types-abstract} and are tied to the compiler's internal operator-handling as described in the following sections.
 
 \subsection{Numeric types}
 \label{types-numeric-types}
 
 \define[Type]{Float}{define-float}{Represents a double-precision IEEE 64-bit floating point number.}
 
 \define[Type]{Int}{define-int}{Represents an integral number.}
-While every \type{Int} can be used where a \type{Float} is expected (that is, \type{Int} \emph{is assignable to} or \emph{unifies with} \type{Float}), the reverse is not true: Assigning a \type{Float} to an \type{Int} might lose precision and is not allowed implicitly.
+While every \type{Int} can be used where a \type{Float} is expected (that is, \type{Int} \emph{is assignable to} or \emph{unifies with} \type{Float}), the reverse is not true: assigning a \type{Float} to an \type{Int} might cause loss of precision and is therefore not implicitly allowed.
```
Simn: implicitly allowed seems worse than allowed implictly here. But maybe I just can't English.

HaxeManual/02-types.tex
```diff
@@ -11,26 +11,26 @@ \chapter{Types}
...
 The Haxe type system knows seven type groups:
 
 \begin{description}
  \item[\emph{Class instance}:] an object of a given class or interface
- \item[\emph{Enum instance}:] a value of a Haxe enumeration
+ \item[\emph{Enum instance}:] a Haxe enumeration
  \item[\emph{Structure}:] an anonymous structure, i.e. a collection of named fields
- \item[\emph{Function}:] a compound type of several arguments and one return
- \item[\emph{Dynamic}:] a wildcard type which is compatible with any type
+ \item[\emph{Function}:] a compound type of several arguments and one return value
```
Simn: This isn't quite accurate still. See that other issue opened recently.

HaxeManual/02-types.tex
```diff
@@ -46,27 +46,27 @@ \section{Basic Types}
...
+For performance reasons, the Haxe Compiler does not enforce any overflow behavior. The burden of checking for numeric overflow falls to the target platform. Here are some platform specific notes on overflow behavior:
 
 \begin{description}
  \item[C++, Java, C\#, Neko, Flash:] 32-bit signed integers with usual overflow practices 
- \item[PHP, JS, Flash 8:] No native \emph{Int} type, loss of precision will occur if they reach their float limit (2\textsuperscript{52})
+ \item[PHP, JS, Flash 8:] No native \emph{Int} type, loss of precision will occur if a number reaches the float limit (2\textsuperscript{52})
 \end{description}
 
-Alternatively, the \emph{haxe.Int32} and \emph{haxe.Int64} classes can be used to ensure correct overflow behavior regardless of the platform at the cost of additional computations depending on the platform.
+If correct overflow behaviour is required, the \emph{haxe.Int32} and \emph{haxe.Int64} classes can be used although this comes at the cost of additional computations depending upon the platform.
```
Simn: We should AE instead of BE.

HaxeManual/02-types.tex
paul59: I don't think it matters; you could make the same argument about 'not therefore' versus 'therefore not'.

HaxeManual/02-types.tex
paul59: Oops. Try this:

"Alternatively, the Int32 and Int64 classes can be used to ensure correct overflow behavior although additional computations may be required on certain platforms."

I don't know whether 'all platforms' or 'certain platforms' is technically correct, I am new to Haxe!

Aurel300:
Added in bacf825.

PR #371, opened 2019-02-20
Correct small typo

haxiomic:
% ~have~ **has** lower precedence than * and / whereas in C these three have the same


haxiomic:
Closing as this is already addressed in #368

PR #376, opened 2019-03-06
Fix outdated reference to -D analyzer

Gama11:
Also removed a stray backtick that seemed to break the markdown on the page:

### DIFF
--- a/HaxeManual/08-compiler-features.tex
+++ b/HaxeManual/08-compiler-features.tex
@@ -725,12 +725,12 @@ class D { } // Type-level disable
 \paragraph{Modules}
 \label{cr-static-analyzer-modules}
 
-The static analyzer currently comes with the following modules. Unless noted otherwise they are enabled if \ic{-D analyzer} is used.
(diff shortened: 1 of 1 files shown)


markknol:
@simn since which version does it work like this? Maybe we mention this change somewhere too?

Gama11:
I think it's been that way since at least 3.3 / 3.4? So not sure if it's worth documenting the previous define for versions older than that at this point.

markknol:
According this blogpost: https://haxe.org/blog/haxe-3.4-release/
> In Haxe 3.4 the static analyzer has settled and runs by default, so the -D analyzer flag has been removed. To have extra optimizations the -D analyzer-optimize can be used. This builds a control flow graph and then the optimizer (if enabled) does some optimizations on that, like folding expressions, removing dead code, etc. This optimization flag is not enabled by default because of too many vars being eliminated for hxcpp, in Haxe 4.0 it will be enabled by default. 

If that is the case, then the copy is not completely correct and first line should be:

```latex
The static analyzer is enabled by default and currently comes with the following modules.  To have extra optimizations \ic{-D analyzer-optimize} can be used. 
```
@Simn

Aurel300:
Added in f5b02db.

Issue #324, opened 2018-02-12
Haxe 4 documentation


markknol:
We have to start making docs for when Haxe 4 is ready. This is a WIP of stuff the needs to be written/documented in the manual.

* [ ] ml plugin docs / example https://github.com/HaxeFoundation/haxe/blob/development/std/eval/vm/Context.hx#L36
* [ ] arrow function syntax
* [ ] function type syntax

PR #383, opened 2019-04-15
Haxe 4 changes

Aurel300:
Features and changes to document, based on the changelogs from the 4.0.0 previews and release candidates:

## Syntax

 - [x] arrow functions
 - [x] `final` keyword, including in anonymous structures
 - [x] `case var x` in `switch`
 - [x] new function type syntax
 - [x] `enum abstract` (replaces `@:enum`)
 - [x] `extern` keyword (replaces `@:extern`)
 - [x] metadata including dots
 - [x] `?` before field name (same as `@:optional`)
 - [x] `Type1 & Type2` for `typedefs` and replacing `(Type1, Type2)` type constraint
 - [x] `for (key => value in e)`
 - [x] `inline call()` and `inline new`
 - [x] reserved `operator` and `overload` keywords
 - [x] `[]` is also an empty map
 - [x] `get`, `set` in property access, `get_x` and `set_x` no longer supported
 - [ ] XML literals

## Stdlib

 - [x] `List` moved to `haxe.ds`
 - [ ] `implements Dynamic` on `extern`s only
 - [ ] `haxe.iterators`
 - [ ] `sys.thread`

## Compiler defines

 - [ ] `warn_var_shadowing`

## Target changes

 - `*.Syntax` classes to replace `untyped __{js,php,...}__`
 - JS
   - [ ] externs for `Map`, `Set`, `Date` ...
   - [ ] ES6 generation via `-D js-es=6`
   - [ ] some externs moved to `js.lib`
 - PHP
   - [ ] `--php-{prefix,front,lib}` changed to `-D php-*=`
   - [ ] minimum version supported is now PHP 7.0
 - Lua
   - [ ] `-D lua-vanilla`
 - macro
   - [ ] `eval` target ?
   - [ ] debugger
   - [ ] `@:persistent` (replacing `registerModuleReuseCall` and `onMacroContextReused`)

## External libraries

 - [ ] `hx3compat`
   - `haxe.remoting`
   - `haxe.web.Request`
   - `haxe.unit`
 - [ ] `record-macros`
 - jQuery and swfobject ?

## Compiler features

 - [x] loop unrolling
 - [ ] null safety

## Other changes

 - [x] `@:op(a.b)`
 - [ ] Unicode support
 - [ ] `@:using`
 - [ ] mention `display` + IDE integration, `vshaxe` ?
 - [ ] enum values without arguments as default function argument values
 - [ ] disallow macro-in-macro


Simn:
Great list!

I'm not sure if I like the workflow with this draft PR, but we can try it.

Aurel300:
I guess ideally there would be many separate PRs for this, but the activity here is low anyway. I made it a branch on this repo and not a fork in case other people contribute. Also I find it nicer to have the checklist in a draft PR rather than a separate issue.

Simn:
Yeah I'm mostly concerned about reviewing this.

HaxeManual/02-types.tex
```diff
@@ -940,6 +940,12 @@ \subsection{Enum abstracts}
 	\item Defining fields requires less syntax.
 \end{itemize}
 
+\since{4.0.0}
+
+Since Haxe 4, it is possible to declare enum abstracts without using the \expr{:enum} metadata, instead using the more natural syntax \expr{enum abstract}. Additionally, if the underlying type is \expr{String} or \expr{Int}, the values for the enum cases can be omitted and are deduced by the compiler. For \expr{Int} abstracts, the deduced values increment the last user-defined value, or start at zero if no value was declared yet:
```
Simn: The example shows both Int and String abstracts, but this paragraph only mentions what the compiler does with Int abstracts.

Aurel300: The code example shows the String values in comments next to the cases. Should I add it in the text as well?

Simn: Well that's my point. The intro text with the colon makes it sound like the example is specific to Int, but then it also shows String. Would be better to quickly mention the latter too.

markknol:
Related https://github.com/HaxeFoundation/HaxeManual/issues/324

HaxeManual/assets/Extension3.hx
```diff
@@ -0,0 +1,15 @@
+typedef PointX = {
+  var x:Int;
+}
+
+typedef PointY = {
+  var y:Int;
+}
+
+typedef PointXY = PointX & PointY;
```
Gama11: I wonder if this is a good example, doesn't seem like something you would do in real code? Maybe something like `typedef Point3D = Point2D & {z: Int}` would make more sense.

HaxeManual/02-types.tex
```diff
@@ -319,6 +319,14 @@ \subsection{Inheritance}
 
 A class may override \tref{methods}{class-field-method} of its parent class, which requires the explicit \expr{override} keyword. The effects and restrictions of this are detailed in \Fullref{class-field-overriding}.
 
+\since{4.0.0}
+
+Classes can be declared with the keyword \expr{final} to prevent them from being extended.
+
+In some situations the user may still want to extend a class marked as final. This is possible by using the \expr{:hack} metadata on the \emph{extending} class. This should be avoided in general, hence the discouraging metadata name \expr{:hack}.
```
Simn: I'd honestly not mention this at all...

Aurel300: I guess it's already listed in the metadata list, but that's a bit hard to navigate … I'll remove it.

nadako: Not sure that even works everywhere, if the generator writes a native `final` flag.

HaxeManual/04-class-field.tex
```diff
@@ -286,7 +286,7 @@ \subsection{Inline}
...
 
-Inline is not guaranteed to be done. The compiler might cancel inlining for various reasons or a user could supply the \ic{--no-inline} command line argument to disable inlining. The only exception is if the class is \tref{extern}{lf-externs} or if the class field has the \expr{:extern} \tref{metadata}{lf-metadata}, in which case inline is forced. If it cannot be done, the compiler emits an error.
+Inline is not guaranteed to be done. The compiler might cancel inlining for various reasons or a user could supply the \ic{--no-inline} command line argument to disable inlining. The only exception is if the class is \tref{extern}{lf-externs} or if the class field has the \tref{\expr{extern}}{class-field-extern} access modifier, in which case inline is forced. If it cannot be done, the compiler emits an error.
```
Simn: Actually, Haxe 4 doesn't cancel inlining anymore at all. The last case where this was done was for macro-in-macro situations, but those are gone now.

HaxeManual/04-class-field.tex
```diff
@@ -323,3 +323,30 @@ \subsection{Static}
...
+\subsection{Final}
+\label{class-field-final}
+
+\since{4.0.0}
+
+The \expr{final} keyword can be used on class fields with the following effects:
+
+\begin{itemize}
+	\item \expr{final function ...} to make a function non-overridable in subclasses.
+	\item \expr{final x = ...} to declare a constant that requires immediate initialization and cannot be written to.
```
Simn: It's immediate or in the constructor, right?

HaxeManual/04-class-field.tex
```diff
@@ -323,3 +323,30 @@ \subsection{Static}
...
+	\item \expr{inline final x = ...} is the same but \tref{inlines}{class-field-inline} the value wherever it is used. Only constant values can be assigned.
```
Simn: This should probably reference what we mean by "constant values".

HaxeManual/04-class-field.tex
```diff
@@ -323,3 +323,30 @@ \subsection{Static}
...
+\end{itemize}
+
+\expr{static final} fields should be initialized immediately by providing an expression. If a class has non-static \expr{final} variables which are not initialized immediately, it requires a constructor which has to assign values to all such fields. \expr{final} does not affect \tref{visibility}{class-field-visibility} and it is not supported on \tref{properties}{class-field-property}.
```
Simn: "should" is too soft here, it's "must".

HaxeManual/04-class-field.tex
Aurel300: I originally had a `\tref` to the constants page. But the current constants page is not constants, it's literals. So we don't have actually have a constants page yet.

HaxeManual/05-expressions.tex
```diff
@@ -186,6 +189,15 @@ \section{Array Declaration}
...
 Advanced initialization techniques are described in \Fullref{lf-array-comprehension}.
 
+\section{Map Declaration}
+\label{expression-map-declaration}
+
+Maps are initialized similarly to arrays, but include both keys and their corresponding values. \expr{["example" => 1, "data" => 2]} initializes a map (\type{Map<String, Int>} specifically) where the key \expr{"example"} stores the value \expr{1} and \expr{"data"} stores the value 2.
+
+\since{4.0.0}
+
+Where a map type is expected, \expr{[]} refers to an empty map.
```
Simn: This could reference the top-down inference section.

HaxeManual/05-expressions.tex
```diff
@@ -272,6 +284,16 @@ \section{var}
 
 The scoping behavior of local variables is described in \Fullref{expression-block}.
 
+\since{4.0.0}
+
+In Haxe 4, the alternative keyword \expr{final} was introduced at the expression level. Variables declared with \expr{final} instead of \expr{var} can only be assigned a value once.
+
+\haxe{assets/Final.hx}
+
+It is important to note that \expr{final} may not have the intended effect with types that are not imutable, such as arrays or objects. Even though the variable cannot have a different object assigned to it, the object itself can still be modified using its methods:
```
Simn: typo "immutable"

HaxeManual/04-class-field.tex
Gama11: So is `extern inline` pointless / redundant now?

Aurel300: What's the point of `extern inline` then?

HaxeManual/06-language-features.tex
```diff
@@ -391,12 +399,20 @@ \subsection{Variable capture}
...
 
 Here, \expr{leafNode} is bound to \expr{Leaf("foo")} if the input matches that. In all other cases, \expr{myTree} itself is returned: \expr{case x} works similar to \expr{case _} in that it matches anything, but with an identifier name like \expr{x} it also binds the matched value to that variable.
 
+\since{4.0.0}
+
+If the compiler detects that the name of a ``catch-all'' variable is very similar to an enum constructor, it will warn about a potential typo. This can be avoided by prefixing the variable identifier with \expr{var}:
+
+\haxe[firstline=107,lastline=113]{assets/PatternMatching.hx}
+
+Any capture variable can be prefixed with \expr{var}, but in most cases it is unnecessary. Since it also circumvents the warning about potential typos, it is recommended to not use \expr{var} unless necessary.
```
Simn: That a weird recommendation. If you explicitly use `var` you probably want to capture a var.

HaxeManual/08-compiler-features.tex
```diff
@@ -734,3 +734,28 @@ \section{Static Analyzer}
 \item[\ic{fusion}]: Moves variable expressions to its usage in case of single-occurrence. By default, only compiler-generated variables are handled. This can be changed by using the compiler flag `\ic{-D analyzer-user-var-fusion} or the metadata \ic{@:analyzer(user_var_fusion)}.
 \item[\ic{purity_inference}]: Infers if fields are ``pure'', i.e. do not have any side-effects. This can improve the effect of the fusion module.
 \end{description}
+
+
+\section{Loop unrolling}
+\label{cr-loop-unrolling}
+\since{4.0.0}
+
+The compiler tries to unroll short iterations of \expr{for} loops over constant ranges. This can produce faster code that uses fewer variables.
```
Simn: Not sure if "fewer variables" is the significant thing to mention here. The main point is that there's no longer a loop at all.

HaxeManual/06-language-features.tex
Aurel300: I'm trying to say you shouldn't use `var` by default for capture variables. That way if you make a typo you are warned. If you get used to always capturing with `var` you might not get a useful warning.

HaxeManual/08-compiler-features.tex
```diff
@@ -734,3 +734,28 @@ \section{Static Analyzer}
...
+The above code turns into something similar to:
+
+\begin{lstlisting}
+test(0);
+test(1);
+test(2);
+test(3);
+\end{lstlisting}
+
+Whether a loop is unrolled or not depends on the number of iterations of the loop and the number of expressions inside the loop body. This threshold can be configured using the \expr{-D loop_unroll_max_cost=...} compile-time define.
```
Simn: This should mention the default value (250) so people get a rough idea of what dimension we're talking about here. Might also want to mention that we multiply the number of iterations with the number of expressions in the loop body.

Simn:
I went through the changes and commented. Could we address these remarks and then merge the PR? That way I don't have to review these parts again later because this is turning into a rather substantial PR.

HaxeManual/04-class-field.tex
Simn: Hmm, I forgot about constructor inlining which is a bit of a separate case. Here the inlining might still be canceled after all and then `extern inline` causes an error. So maybe it's alright to keep this as-is.

HaxeManual/06-language-features.tex
Simn: I don't get it, why would you want a warning if you want to capture a variable anyway? In that case it doesn't matter how you name it and how similar the name is to an enum constructor.

Aurel300: Never mind, you are right. I was thinking about it the wrong way.

HaxeManual/04-class-field.tex
Aurel300: #386

PR #394, opened 2019-05-29
Fix function name

RblSb:


Gama11:
What's wrong with the old name?

RblSb:
This is recursive, or i miss something?

RblSb:
Oops, `this` is not current place for abstract, confused with class `this`.

PR #395, opened 2019-05-30
Auto-generated define / metadata tables

Aurel300:
Committed generated tables - for now JSONs taken from the updated https://github.com/Aurel300/haxe/tree/feature/manual-references. Also updated
 `README.md` and `Makefile` so the process is less annoying.


generate/generate.hxml
```diff
@@ -0,0 +1,2 @@
+--main Generate
+--neko generate.n
```
Gama11: `--interp` might be better?

PR #401, opened 2019-06-12
Update 05-expressions.tex

RichardBray:
Coming from Javascript I didn't know what a `cast` was. This definition will help others in my position.

### DIFF
--- a/HaxeManual/05-expressions.tex
+++ b/HaxeManual/05-expressions.tex
@@ -615,6 +615,7 @@ A value which is thrown like this can be caught by \tref{\expr{catch} blocks}{ex
 \section{cast}
 \label{expression-cast}
 
+The `cast` keyword changes the percieved type of an expression without changing it's contents.
 Haxe allows two kinds of casts:
 
 \begin{lstlisting}
--- a/HaxeManual/05-expressions.tex
+++ b/HaxeManual/05-expressions.tex
@@ -615,7 +615,7 @@ A value which is thrown like this can be caught by \tref{\expr{catch} blocks}{ex
 \section{cast}
 \label{expression-cast}
 
-The `cast` keyword changes the percieved type of an expression without changing it's contents.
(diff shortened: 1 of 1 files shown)


HaxeManual/05-expressions.tex
```diff
@@ -615,6 +615,7 @@ \section{throw}
 \section{cast}
 \label{expression-cast}
 
+The `cast` keyword changes the perceived type of an expression without changing its contents.
```
Aurel300: Please use `\expr{cast}` instead of Markdown syntax. I'm also not sure the definition is strictly correct… Perhaps "reinterprets" instead of "changes"?

Simn:
This raises more questions than it answers. What does "reinterpret" mean? Who perceives this? What are "contents"?

I don't really see why we need any additional information in this chapter in the first place. The details of casts are described in the two sub-chapters. Trying to somehow summarize this is only going to lead to additional confusion, especially because unsafe and safe casts are fundamentally different.

What we could do is remove this chapter entirely so the sub-chapters are promoted. That makes sense for multiple reasons and avoids the problem of defining what a cast is in general.

RichardBray:
@Simn I personally believe for those new to Haxe/Programming who have no idea what the concept of casting is a short definition would be beneficial as not many other languages use the `cast` keyword. Maybe this would be better suited to the Haxe cookbook --> https://code.haxe.org/category/beginner/index.html

Simn:
I get that, the issue with casts is that they are quite intricate and any attempt to summarize them in one sentence will either be mundane or confusing. We should instead improve the beginning of the unsafe-cast/safe-cast chapters directly in a way that even JavaScript programmers get an idea what's going on.

And yes a more elaborate example would be a good fit for the cookbook and could be linked from the manual.

RichardBray:
@Simn Okay that makes sense, I will put this definition to one side and think about ways to improve the current pages.  Thanks for your help.

PR #402, opened 2019-06-13
Operators

### LINKED ISSUE #238: Properly document operator precedence
https://github.com/HaxeFoundation/haxe/issues/4121

### LINKED ISSUE #372: Operator precedence
On the following line of documentation:

https://github.com/HaxeFoundation/HaxeManual/blob/7f11818fabd74a0c6773c068314588412e96a3c1/HaxeManual/02-types.tex#L142

    % have lower precedence than * and / whereas in C these three have the same

I may be misunderstanding the definition of the word "precedence" here. My understanding is that lower precedence means that the operator would be executed _later_ in the sequence of expression resolution. In [this issue](https://github.com/HaxeFoundation/haxe/issues/7070) on the main Haxe repository, it seems as though modulo has a higher precedence according to this definition. Indeed, modulo is listed above multiply/divide in the table which is stated to be ordered in descending precedence. 

The same could be said for the bitwise operators mentioned two lines below (they are _higher_, not _lower_ precedence than comparison operators).

Am I misunderstanding the documentation? If not, I will go ahead and submit a PR fixing this issue.

### LINKED ISSUE #385: Operators
- [ ] specify operator precedence clearly (#372)
 - [ ] move operators into their proper place in the Expressions chapter

Aurel300:
Closes #385, closes #372, closes #238.

Todo:

 - precedence of `cast`?
 - precedence of ECheckType?

### DIFF
--- a/HaxeManual/05-expressions.tex
+++ b/HaxeManual/05-expressions.tex
@@ -169,14 +169,11 @@ The Haxe syntax supports the following constants:
 Furthermore, the internal syntax structure treats \tref{identifiers}{define-identifier} as constants, which may be relevant when working with \tref{macros}{macro}.
 
 
-\section{Binary Operators}
-\label{expression-binops}
+\section{Operators}
+\label{epxression-operators}
 
-
-
-
-\section{Unary Operators}
-\label{expression-unops}
+\subsection{Unary Operators}
+\label{expression-operators-unops}
 
 \begin{center}
 \begin{tabular}{| l | l | l | l | l |}
@@ -190,6 +187,20 @@ Furthermore, the internal syntax structure treats \tref{identifiers}{define-iden
 \end{tabular}
 \end{center}
 
+
+\subsection{Ternary Operator}
+\label{expression-operators-ternary}
+
+Haxe supports the ternary conditional operator \expr{cond ? whenTrue : whenFalse}:
+
+\begin{lstlisting}
+trace(true ? "Haxe" : "Neko"); // Haxe
+trace(1 == 2 ? 3 : 4); // 4
+\end{lstlisting}
+
+
+\paragraph{Increment and decrement}
+
 The increment and decrement operators change the value of a variable, so they cannot be used on a constant. They also produce different results based on whether they are used as a prefix operator (change value then return it) or as a postfix operator (return value then change it).
 
 \begin{lstlisting}
@@ -202,7 +213,15 @@ trace(++a); // 11
 trace(a); // 11
 \end{lstlisting}
 
-The operators specified in the table above specify the types and meanings for operations on basic types. Additional functionality can be implemented using \tref{abstract operator overloading}{types-abstract-operator-overloading} or \tref{macro processing}{macro}.
+\subsection{Precedence}
+\label{expression-operators-precedence}
+
+All \tref{unary operators}{expression-operators-unops} have a higher precedence than the \tref{binary operators}{expression-operators-binops}.
+
+\subsection{Overloading and macros}
+\label{expression-operators-overloading}
+
+The operators specified in the previous sections specify the types and meanings for operations on basic types. Additional functionality can be implemented using \tref{abstract operator overloading}{types-abstract-operator-overloading} or \tref{macro processing}{macro}.
 
 For macro processing in particular, there is an additional operator available: the postfix \expr{!} operator.
 
--- a/HaxeManual/05-expressions.tex
+++ b/HaxeManual/05-expressions.tex
@@ -238,6 +238,17 @@ In Haxe, the result of a modulo operation always keeps the sign of the dividend 
 \end{tabular}
 \end{center}
 
+\paragraph{Logical operators}
+
+\begin{center}
+\begin{tabular}{| l | l | l | l | l |}
+	\hline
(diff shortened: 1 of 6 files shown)


HaxeManual/05-expressions.tex
```diff
@@ -169,11 +169,242 @@ \section{Constants}
...
+	\expr{<<=} & shift left & \type{Int} & \type{Int} & \type{Int} \\
+	\expr{>>=} & shift right & \type{Int} & \type{Int} & \type{Int} \\
+	\expr{>>>=} & unsigned shift right & \type{Int} & \type{Int} & \type{Int} \\
+	\expr{\&=} & bitwise and & \type{Int} & \type{Int} & \type{Int} \\
+	\expr{|=} & bitwise or & \type{Int} & \type{Int} & \type{Int} \\
+	\expr{\^=} & bitwise xor & \type{Int} & \type{Int} & \type{Int}
+\end{tabular}
+\end{center}
+
+In all cases, a compound assignment modifies the value of a variable, then returns the new value.
```
Simn: I know you can interpret fields and arrays as "variable", but I'd still make it clear that we're not strictly talking about `var`s here.

Simn: Also not sure about the "returns" terminology. I think "evaluates to" is better to avoid potential confusion.

Simn: Another thing that could be mentioned is that you can't do `Int += Float`, which is different from `Int + Float` where this works just fine.

HaxeManual/05-expressions.tex
```diff
@@ -169,11 +169,242 @@ \section{Constants}
...
+	\expr{==} & equal & \type{String} & \type{String} & \type{Bool} \\
+	\expr{!=} & not equal & \type{String} & \type{String} & \type{Bool} \\
+	\expr{<} & lexicographically before & \type{String} & \type{String} & \type{Bool} \\
+	\expr{<=} & lexicographically before or equal & \type{String} & \type{String} & \type{Bool} \\
+	\expr{>} & lexicographically after & \type{String} & \type{String} & \type{Bool} \\
+	\expr{>=} & lexicographically after or equal & \type{String} & \type{String} & \type{Bool}
+\end{tabular}
+\end{center}
+
+\type{String} is an immutable type in Haxe, so two separate instances of the same string are considered equal.
```
Simn: That's not really related to each other because in general run-times don't ensure that memory for equal string values is always shared.

HaxeManual/05-expressions.tex
```diff
@@ -169,11 +169,242 @@ \section{Constants}
...
+
+\paragraph{Equality operators}
+
+\begin{center}
+\begin{tabular}{| l | l | l | l | l |}
+	\hline
+	Operator & Operation & Operand 1 & Operand 2 & Result type \\ \hline
+	\expr{==} & equal & any & same as operand 1 & \type{Bool} \\
```
Simn: "same" is not accurate, it's fine as long as the assignment can be made. For instance `baseClass == childClass` works as expected.

HaxeManual/05-expressions.tex
```diff
@@ -169,11 +169,242 @@ \section{Constants}
...
+\emph{For enums:}
+
+\begin{description}
+	\item[Enums without parameters] always represent the same value, so \expr{MyEnum.A == MyEnum.A}. 
+	\item[Enums with parameters] can be compared with \expr{a.equals(b)} (which is short for \expr{Type.enumEquals()}).
+\end{description}
+
+\emph{Dynamic:}
+
+Comparison involving at least one \type{Dynamic} value is unspecified and platform-specific.
```
Simn: I'd stay away from the term "value" here. What's a "Dynamic value"? The general problem is about comparison of unrelated types, through `Dynamic` or some other mechanism.

HaxeManual/05-expressions.tex
```diff
@@ -169,11 +169,242 @@ \section{Constants}
...
+	\expr{...} & interval (see \tref{range iteration}{expression-for}) & \type{Int} & \type{Int} & \type{IntIterator} \\
+	\expr{=>} & arrow (see \tref{map}{expression-map-declaration}, \tref{key-value iteration}{expression-for}, \tref{map comprehension}{lf-map-comprehension}) & any & any & -
+\end{tabular}
+\end{center}
+
+
+\subsection{Ternary Operator}
+\label{expression-operators-ternary}
+
+Haxe supports the ternary conditional operator \expr{cond ? whenTrue : whenFalse}:
```
Simn: Since we're listing types everywhere else, it would be nice to mention the types for this as well.

HaxeManual/05-expressions.tex
Aurel300: Is that what the current wording implies? "two separate instances" – i.e. not the same in memory – "are _considered_ equal" – i.e. by the Haxe `==` operator. What should I clarify?

HaxeManual/05-expressions.tex
Aurel300: How about `any & any` in the table and a note saying that the types of operands 1 and 2 must unify (+ link to unification section)?

HaxeManual/05-expressions.tex
Aurel300: I rewrote it and added an example.

PR #393, opened 2019-05-29
typo fix in HaxeManual/02-types.tex

iamrajiv:

PR #404, opened 2019-06-16
small typo fix in 02-types.tex

iamrajiv:


Aurel300:
As far as I can see, all the changes you are introducing are incorrect and have already been addressed in your [previous `02-types.tex` PR](https://github.com/HaxeFoundation/HaxeManual/pull/393#discussion_r288489767). I appreciate your help with the manual, but these are not very substantial changes – you could put changes in many files into a single PR. Please make sure you do not reintroduce the same changes as before, especially if they were previously declined.

PR #408, opened 2019-06-18
Remove custom commands

### LINKED ISSUE #407: Remove custom LaTeX commands
The manual has a couple of custom LaTeX commands:

 - `\todo` - generates the `todo.txt` file, which is seriously out of date and I don't think it should ever be committed (again); superseded by GH issues (if we can be bothered to open them) or even a project board
 - `\state` - indicates whether a section has been modified, reviewed, or has no content; the latter is obvious and should be an open issue if it is an important part (e.g. the missing target-specific sections); the former is superseded by reviews of PRs
 - `\flag` - `\flag{fold}{true}` is the only use of this, in the [resources section](https://haxe.org/manual/cr-resources.html) where it embeds the subsections directly into the section; we can use `\paragraph` for this
 - `\maintainer` - never used
 - `\subtoc` - never used
 - `\todototoc` - never used, no-op in the parser
 - `\listoftodos` - never used, no-op in the parser

I think we should get rid of all of the above.

Aurel300:
Closes #407.

Removed: `\todo`, `\flag`, `\state`, `\maintainer`, `\multicolumn`, `\todotoc`, `\listoftodos`, `\subtoc`

Simplified parser a bit (remove no-op LaTeX commands).

Here is what the TODO list looked like for future reference:

 - [ ] 1.1 - What is Haxe?: Could we have a big Haxe logo in the First Manual Page (Introduction) under the menu (a bit like a book cover ?) It looks a bit empty now and is a landing page for "Manual"
 - [ ] 1.3 - Hello World: This generates the following output: too many 'this'. You may like a passive sentence: the following output will be generated...though this is to be avoided, generally
 - [ ] 2.1.4 - Void: please review, doubled content
 - [ ] 2.1.4 - Void: review please, sounds weird
 - [ ] 2.2 - Nullability: for starters...please review
 - [ ] 2.2.1 - Optional Arguments and Nullability: Is there a difference between `?y : Int` and `y : Null<Int>` or can you even do the latter? Some more explanation and examples with native optional and Haxe optional arguments and how they relate to nullability would be nice.
 - [ ] 2.3 - Class Instance: please review future tense
 - [ ] 2.4 - Enum Instance: Same as in 2.2, what is `Enum<T>` syntax?
 - [ ] 2.4.1 - Enum Constructor: list arguments
 - [ ] 2.5 - Anonymous Structure: please reformat
 - [ ] 2.6 - Function Type: It seems a bit convoluted explanations. Should we maybe start by "decoding" the meaning of  Void -> Void, then Int -> Bool -> Float, then maybe have samples using $type
 - [ ] 2.8.1 - Implicit Casts: please review your use of "this" and try to vary somewhat to avoid too much word repetition
 - [ ] 2.8.2 - Operator Overloading: please review for correctness
 - [ ] 2.8.3 - Array Access: You have marked "Map" for some reason
 - [ ] 3.5 - Unification: Mention toString()/String conversion somewhere in this chapter.
 - [ ] 3.5.1 - Between Class/Interface: "parent class" should probably be used here, but I have no idea what it means, so I will refrain from changing it myself.
 - [ ] 3.7.4 - Resolution Order: proper label and caption + code/identifier styling for diagram
 - [ ] 6.4.10 - Exhaustiveness checks: Figure out wtf our rules are now for when this is checked.
 - [ ] 9.5.1 - Enum building: Check if we can build GADTs this way.
 - [ ] 10.4.3 - Integer Math: I think C++ can use integer operations, but I don't know about any other targets. Only saw this mentioned in an old discussion thread, still true?

Missing Content:
 - [x] 5.4 - Operators
 - [ ] 9.6.4 - Type Parameters
 - [ ] 10.11 - Input/Output
 - [ ] 12.4 - Neko
 - [ ] 12.6.5 - Using C++ Pointers
 - [ ] 12.10 - Python
 - [ ] 12.12 - HashLink


Aurel300:
(Obsoleted by #409)

PR #411, opened 2019-06-27
01-intro, target code generated, not Haxe code

uvtc:


uvtc:
... unless I'm misunderstanding what you mean by generating code.

Simn:
It never occurred to me to read it like that, but now that you mention it I suppose your version is better. My thinking was "take the Haxe code and generate it", but that's not really accurate.

PR #412, opened 2019-06-27
07-compiler-usage, s/--cp/--class-path/

uvtc:
The output of `haxe --help` lists only `-p` and `--class-path`.


Aurel300:
Thanks, I was too eager to replace all `-` switches with `--` it seems!

PR #419, opened 2019-07-16
Fix wrong example code used for Array pattern matching

R2RT:
Somehow Struct pattern matching sample code got duplicated here after splitting done in 7de2ed9.
Original code was retrieved and reformatted.

I did not bother to create corresponding issue.


Aurel300:
Thanks, well spotted!

PR #423, opened 2019-08-01
Add the syntax of how to type functions

CanIGetaPR:
I was reading this document to learn how to declare a function signature when creating a variable, and this document lacked an example of that.

### DIFF
--- a/content/02-types.md
+++ b/content/02-types.md
@@ -571,7 +571,7 @@ As shown in the above example, the type of a function which takes no arguments u
 
 ##### Old function type notation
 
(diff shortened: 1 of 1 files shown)


Simn:
This is not about how you as the author types the code, but about the compiler typing the code you wrote. Maybe we can make this a bit clearer, but this change isn't quite working in that regard.

CanIGetaPR:
Yes, I'm not exactly sure how to do that but I had to ask someone for help on how to define a variable function after reading the readme here so it definitely lacks some information but I'm not sure how to add it.

Simn:
Closing this as we want to use the new function type syntax anyway.

PR #428, opened 2019-08-16
Hello World: fix problem in haxe command

nikolas:
`haxe --main` isn't a valid parameter. Looks like it should be `haxe -main`.


Aurel300:
It is a valid parameter starting from Haxe 4. Unfortunately I made these changes consistently across all of the manual already … Maybe we should have a note about this somewhere.

PR #429, opened 2019-08-16
fix compile.hxml syntax

larsiusprime:
There ought only to be only "-" instead of "--" here, right? I just tried it myself and it only works with the single-dash syntax when using compile.hxml rather than from the command line. Am I missing something?


Aurel300:
#428 

I should never have pushed these changes yet x)

Simn:
Hmm, this is a bit bad while Haxe 4 isn't out yet...

Aurel300:
Ok, hopefully I didn't miss any in a67addd7c10d784e1da837ef365598b9a89f3174

PR #431, opened 2019-09-01
Change invalid function name

judehunter:
An iterable class with a KeyValueIterator must implement `keyValueIterator() : KeyValueIterator<K,V>` instead of `iterator()`


Simn:
Good catch, thanks!

PR #439, opened 2019-11-22
use `inline final` not `inline var`

nadako:
Please review my engrish.

Also now with `final` in place it's even more awkward that we have the notion of "variable" fields, but I guess changing the terminology now would only bring more confusion.

### DIFF
--- a/content/04-class-field.md
+++ b/content/04-class-field.md
@@ -337,7 +337,7 @@ Note that even though we call such kind of fields "variables", inline variables 
 
 > ##### Trivia: `inline var`
 >
-> Prior to Haxe 4, there was no `final` keyword. The inline variables feature however was present for a long time, using `var` keyword instead of `final`. Using `inline var` still works in Haxe 4 but might be deprecated in the future, because `final` is more appropriate.
+> Prior to Haxe 4, there was no `final` keyword. The inline variables feature however was present for a long time, using the `var` keyword instead of `final`. Using `inline var` still works in Haxe 4 but might be deprecated in the future, because `final` is more appropriate.
 
 <!--label:class-field-dynamic-->
 #### Dynamic
--- a/content/04-class-field.md
+++ b/content/04-class-field.md
@@ -333,16 +333,11 @@ Main.main();
 })({});
 ```
 
-Note that even though we call such kind of fields "variables", inline variables can never
-be reassigned as the value must be known at compile-time to be inlined at the place of usage.
-This makes inline variables a subset of [`final` fields](class-field-final), hence the usage
-of the `final` keyword in the code example above.
+Note that even though we call such kind of fields "variables", inline variables can never be reassigned as the value must be known at compile-time to be inlined at the place of usage. This makes inline variables a subset of [`final` fields](class-field-final), hence the usage of the `final` keyword in the code example above.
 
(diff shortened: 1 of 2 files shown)


content/04-class-field.md
```diff
@@ -333,7 +333,16 @@ Main.main();
...
 ```
 
+Note that even though we call such kind of fields "variables", inline variables can never
+be reassigned as the value must be known at compile-time to be inlined at the place of usage.
+This makes inline variables a subset of [`final` fields](class-field-final), hence the usage
+of the `final` keyword in the code example above.
 
+> ##### Trivia: `inline var`
+>
+> Prior to Haxe 4, there was no `final` keyword, but inline variables feature was present
```
Aurel300: I think this would be better in two sentences – "Prior to … keyword. Inline variables …"

Also, "Inline variables were present …" or "The inline variables feature was present …"

content/04-class-field.md
```diff
@@ -333,7 +333,16 @@ Main.main();
...
+> for a long time, just using `var` keyword instead of `final`. Using `inline var` still
```
Aurel300: "just" can be dropped, I think.

content/04-class-field.md
```diff
@@ -333,7 +333,16 @@ Main.main();
...
+> works in Haxe 4 but might be deprecated in future, because `final` is more appropriate.
```
Aurel300: Apparently "in future" is a thing in British English, but the manual uses American :D ("in the future"?)

content/04-class-field.md
```diff
@@ -333,7 +333,16 @@ Main.main();
 })({});
 ```
 
+Note that even though we call such kind of fields "variables", inline variables can never
```
Aurel300: I don't think text should be manually wrapped, to be consistent with the rest of the manual.

content/04-class-field.md
nadako: Well the point is to say that even though there was no "final" keyword, inline vars were already there.

Aurel300: I understand the intent. Maybe connect the sentences with a `–` rather than `, but`?

nadako:
Fixed the remarks, thanks @Aurel300!

content/04-class-field.md
```diff
@@ -333,16 +333,11 @@ Main.main();
...
-This makes inline variables a subset of [`final` fields](class-field-final), hence the usage
-of the `final` keyword in the code example above.
+Note that even though we call such kind of fields "variables", inline variables can never be reassigned as the value must be known at compile-time to be inlined at the place of usage. This makes inline variables a subset of [`final` fields](class-field-final), hence the usage of the `final` keyword in the code example above.
 
 > ##### Trivia: `inline var`
 >
-> Prior to Haxe 4, there was no `final` keyword, but inline variables feature was present
-> for a long time, just using `var` keyword instead of `final`. Using `inline var` still
-> works in Haxe 4 but might be deprecated in future, because `final` is more appropriate.
+> Prior to Haxe 4, there was no `final` keyword. The inline variables feature however was present for a long time, using `var` keyword instead of `final`. Using `inline var` still works in Haxe 4 but might be deprecated in the future, because `final` is more appropriate.
```
Aurel300: "the `var` keyword"

PR #446, opened 2020-01-13
Proofreading Effort: 02-types.md

ZanaDev:
Ongoing from #445 

Includes a ton of general grammatical adjustments and punctuation corrections. There's a lot, so make sure to check for any errors that I may have introduced.

I'm requesting feedback on two aspects of these changes:
- Should instances of metadata be displayed as "@:metadata" or ":metadata"? I've seen both used, and I'm unsure of why this is the case. For now, I defaulted everything to use the @ prefix since it's actually used in code, but I may be missing the point.
- Should instances of the word "abstract/abstracts" be replaced with "abstract type/abstract types"? I am not certain if the first is completely appropriate, but I have no strong feelings either way.

My overall plan after going through this page has changed slightly:
- I plan to move all Trivia boxes down to the bottom of any respective section they reside in (documentation first, fun stuff second; if a trivia box is holding important information it shouldn't be trivia)
- I will add links, but only for keywords and phrases that should have already been linked (Haxe Standard Library should link to the std page, for instance)

If you have any questions or concerns about how I'm editing the documentation, please let me know! Thanks.

PR #447, opened 2020-01-15
Proofreading Effort: 03-type-system.md

ZanaDev:
Ongoing from #446 

Bet you can't guess what this is about. Grammar and punctuation changes for everybody. Please take the time to point out any errors I've introduced.

I'm also requesting that someone who knows what they're talking about rewrite line 6 on this page. The current information on that line seems to be referring to an example that doesn't exist, or is just written in such a confusing manner that I can't actually really understand the actual intent of it without reading the page it links. I think it should be written in a more introductory, simpler manner than it is now.

If you have any questions or concerns let me know, thanks.

### DIFF
--- a/content/03-type-system.md
+++ b/content/03-type-system.md
@@ -1,23 +1,23 @@
 <!--label:type-system-->
 ## Type System
 
-We learned about the different kinds of types in [Types](types) and it is now time to see how they interact with each other. We start off easy by introducing [typedef](type-system-typedef), a mechanism to give a name (or alias) to a more complex type. Among other things, this will come in handy when working with types having [type parameters](type-system-type-parameters).
+We learned about the different kinds of types in [Types](types) and it is now time to see how they interact with each other. We start off easy by introducing [typedef](type-system-typedef), a mechanism to give a name (or alias) to a more complex type. Among other use cases, typedefs will come in handy when working with types that have [type parameters](type-system-type-parameters).
 
-A lot of type-safety is achieved by checking if two given types of the type groups above are compatible. Meaning, the compiler tries to perform **unification** between them as detailed in [Unification](type-system-unification).
+ !!!-REVIEW ME: What is this sentence even referring to? "Type groups" is also unclear in the sense that this phrase has not been used before and may have special meaning of some sort-!!! A lot of type-safety is achieved by checking if two given types of the type groups above are compatible. Meaning, the compiler tries to perform **unification** between them as detailed in [Unification](type-system-unification).
 
 All types are organized in **modules** and can be addressed through **paths**. [Modules and Paths](type-system-modules-and-paths) will give a detailed explanation of the related mechanics.
 
 <!--label:type-system-typedef-->
 ### Typedef
 
-We briefly looked at typedefs while talking about [anonymous structures](types-anonymous-structure) and saw how we could shorten a complex [structure type](types-anonymous-structure) by giving it a name. This is precisely what typedefs are good for. Giving names to structure types might even be considered their primary use. In fact, it is so common that the distinction appears somewhat blurry and many Haxe users consider typedefs to actually **be** the structure.
+We briefly looked at typedefs while talking about [anonymous structures](types-anonymous-structure) and saw how we could shorten a complex [structure type](types-anonymous-structure) by giving it a name. This is precisely why typedefs are useful. Giving names to structure types might even be considered their primary function, and is so common that the distinction between the two appears somewhat blurry. Many Haxe users consider typedefs to actually **be** the structure.
 
 A typedef can give a name to any other type:
 
 ```haxe
 typedef IA = Array<Int>;
 ```
-This enables us to use `IA` in places where we would normally use `Array<Int>`. While this saves only a few keystrokes in this particular case, it can make a much bigger difference for more complex, compound types. Again, this is why typedef and structures seem so connected:
+This enables us to use `IA` in places where we would normally use `Array<Int>`. While this saves only a few keystrokes in this particular case, it can make a larger difference for more complex, compound types. Again, this is why typedef and structures seem so connected:
 
 ```haxe
 typedef User = {
@@ -25,7 +25,7 @@ typedef User = {
   var name : String;
 }
 ```
-A typedef is not a textual replacement but actually a real type. It can even have [type parameters](type-system-type-parameters) as the `Iterable` type from the Haxe Standard Library demonstrates:
+A typedef is not a textual replacement, but actually a real type. They can even have [type parameters](type-system-type-parameters) as the `Iterable` type from the Haxe Standard Library demonstrates:
 
 ```haxe
 typedef Iterable<T> = {
@@ -45,12 +45,12 @@ class Array<T> {
   function push(x : T) : Int;
 }
 ```
-Whenever an instance of `Array` is created, its type parameter `T` becomes a [monomorph](types-monomorph). That is, it can be bound to any type, but only one at a time. This binding can happen
+Whenever an instance of `Array` is created, its type parameter `T` becomes a [monomorph](types-monomorph). That is, it can be bound to any type, but only one at a time. This binding can happen either:
 
-* explicitly by invoking the constructor with explicit types (`new Array<String>()`) or
-* implicitly by [type inference](type-system-type-inference), e.g. when invoking `arrayInstance.push("foo")`.
+* explicitly, by invoking the constructor with explicit types (`new Array<String>()`) or
+* implicitly, by [type inference](type-system-type-inference), e.g. when invoking `arrayInstance.push("foo")`.
 
-Inside the definition of a class with type parameters, these type parameters are an unspecific type. Unless [constraints](type-system-type-parameter-constraints) are added, the compiler has to assume that the type parameters could be used with any type. As a consequence, it is not possible to access fields of type parameters or [cast](expression-cast) to a type parameter type. It is also not possible to create a new instance of a type parameter type, unless the type parameter is [generic](type-system-generic) and constrained accordingly. 
+Inside the definition of a class with type parameters, the type parameters are an unspecific type. Unless [constraints](type-system-type-parameter-constraints) are added, the compiler has to assume that the type parameters could be used with any type. As a consequence, it is not possible to access the fields of type parameters or [cast](expression-cast) to a type parameter type. It is also not possible to create a new instance of a type parameter type unless the type parameter is [generic](type-system-generic) and constrained accordingly. 
 
 The following table shows where type parameters are allowed:
 
@@ -63,15 +63,15 @@ Function | invocation | Allowed for methods and named local lvalue functions.
 Structure | instantiation | 
  
 
-With function type parameters being bound upon invocation, such a type parameter (if unconstrained) accepts any type. However, only one type per invocation is accepted. This can be utilized if a function has multiple arguments:
+As function type parameters are bound upon invocation, they accept any type if left unconstrained. However, only one type per invocation is accepted. This can be utilized if a function has multiple arguments:
 
 [code asset](assets/FunctionTypeParameter.hx)
 
-Both arguments `expected` and `actual` of the `equals` function have type `T`. This implies that for each invocation of `equals` the two arguments must be of the same type. The compiler admits the first call (both arguments being of `Int`) and the second call (both arguments being of `String`) but the third attempt causes a compiler error.
+Both of the `equals` function's arguments, `expected` and `actual`, have type `T`. This implies that for each invocation of `equals`, the two arguments must be of the same type. The compiler permits the first call (both arguments being of `Int`) and the second call (both arguments being of `String`) but the third attempt causes a compiler error due to a type mismatch.
 
 > ##### Trivia: Type parameters in expression syntax
 >
-> We often get the question of why a method with type parameters cannot be called as `method<String>(x)`. The error messages the compiler gives are not very helpful. However, there is a simple reason for that: The above code is parsed as if both `<` and `>` were binary operators, yielding `(method < String) > (x)`.
+> We often get the question of why a method with type parameters cannot be called as `method<String>(x)`. The error messages the compiler gives are not very helpful. However, there is a simple reason for that: the above code is parsed as if both `<` and `>` were binary operators, yielding `(method < String) > (x)`.
 
 <!--label:type-system-type-parameter-constraints-->
 #### Constraints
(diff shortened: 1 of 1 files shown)


content/03-type-system.md
```diff
@@ -1,31 +1,31 @@
...
+We learned about the different kinds of types in [Types](types) and it is now time to see how they interact with each other. We start off easy by introducing [typedef](type-system-typedef), a mechanism to give a name (or alias) to a more complex type. Among other use cases, typedefs will come in handy when working with types that have [type parameters](type-system-type-parameters).
 
-A lot of type-safety is achieved by checking if two given types of the type groups above are compatible. Meaning, the compiler tries to perform **unification** between them as detailed in [Unification](type-system-unification).
+ !!!-REVIEW ME: What is this sentence even referring to? "Type groups" is also unclear in the sense that this phrase has not been used before and may have special meaning of some sort-!!! A lot of type-safety is achieved by checking if two given types of the type groups above are compatible. Meaning, the compiler tries to perform **unification** between them as detailed in [Unification](type-system-unification).
```
RealyUniqueName: I think "of the type groups above" words can be removed. 
Looks like "type groups" means typedefs, abstracts, enums etc.

Simn: Agreed, this can be removed.

content/03-type-system.md
```diff
@@ -1,31 +1,31 @@
...
+This enables us to use `IA` in places where we would normally use `Array<Int>`. While this saves only a few keystrokes in this particular case, it can make a larger difference for more complex, compound types. Again, this is why typedef and structures seem so connected:
 
 ```haxe
 typedef User = {
   var age : Int;
   var name : String;
 }
 ```
-A typedef is not a textual replacement but actually a real type. It can even have [type parameters](type-system-type-parameters) as the `Iterable` type from the Haxe Standard Library demonstrates:
+A typedef is not a textual replacement, but actually a real type. They can even have [type parameters](type-system-type-parameters) as the `Iterable` type from the Haxe Standard Library demonstrates:
```
Simn: Going from singular "A typedef" to plural "they" sounds strange to me. Maybe this paragraph should start plural?

content/03-type-system.md
```diff
@@ -155,16 +155,16 @@ Main.main();
...
+It is not possible to construct normal type parameters; for example, `new T()` would register as a compiler error. The reason for this is that Haxe generates only a single function and the construct would make no sense in that case. This is different when the type parameter is generic: since we know that the compiler will generate a distinct function for each type parameter combination, it is possible to replace the `T` `new T()` with the real type.
 
 [code asset](assets/GenericTypeParameter.hx)
 
-It should be noted that [top-down inference](type-system-top-down-inference) is used here to determine the actual type of `T`. There are two requirements for this kind of type parameter construction to work: The constructed type parameter must be
+It should be noted that [top-down inference](type-system-top-down-inference) is used here to determine the actual type of `T`. There are two requirements for this kind of type parameter construction to work: The constructed type parameter must be:
```
Simn: In other places you're changing it to lower-case-after-colon, so I suppose we should do this here too. I don't like this double-colon in general though, so maybe this can be rephrased differently altogether.

content/03-type-system.md
```diff
@@ -188,21 +188,21 @@ Main.main = function() {
...
 
-If `Array` had no `push()` method and no other means of modification, the assignment would be safe because no incompatible type could be added to it. In Haxe, we can achieve this by restricting the type accordingly using [structural subtyping](type-system-structural-subtyping):
+If `Array` had no `push()` method and no other means of modification, the assignment would be safe as no incompatible type could be added to it. This can be achieved by restricting the type accordingly using [structural subtyping](type-system-structural-subtyping):
 
 [code asset](assets/Variance3.hx)
 
-We can safely assign with `b` being typed as `MyArray<Base>` and `MyArray` only having a `pop()` method. There is no method defined on `MyArray` which could be used to add incompatible types, it is thus said to be **covariant**.
+We can safely assign with `b` being typed as `MyArray<Base>` and `MyArray` only having a `pop()` method. There is no method defined on `MyArray` which could be used to add incompatible types. It is thus said to be **covariant**.
```
Simn: For later: I'm noticing that this entire section can be pretty confusing with regards to variance. It keeps talking about modifications to the array, but that's not _really_ the point here. The significant difference is that the type parameter in question appears in a function argument, which makes the value "writable" from outside. What exactly "push" means has no relevance.

content/03-type-system.md
```diff
@@ -454,18 +454,18 @@ private abstract A { ... }
 
 > ##### Define: Private type
 >
-> A type can be made private by using the `private` modifier. As a result, the type can only be directly accessed from within the [module](define-module) it is defined in.
+> A type can be made private by using the `private` modifier. Afterwards, the type can only be directly accessed from within the [module](define-module) it is defined in.
 > 
-> Private types, unlike public ones, do not become a member of their containing package.
+> Private types, unlike public types, do not become a member of their containing package.
```
Simn: Out of curiosity: Is the previous version wrong or is that just a style thing?

content/03-type-system.md
```diff
@@ -475,41 +475,41 @@ In this example, we are actually importing a **module**, not just a specific typ
 
 [code asset](assets/Import2.hx)
 
-The type `Binop` is an [enum](types-enum-instance) declared in the module `haxe.macro.Expr`, and thus available after the import of said module. If we were to import only a specific type of that module, e.g. `import haxe.macro.Expr.ExprDef`, the program would fail to compile with `Class not found : Binop`.
+The type `Binop` is an [enum](types-enum-instance) declared in the module `haxe.macro.Expr`, and thus available after the import of said module. If we were to import only a specific type of that module, for example, `import haxe.macro.Expr.ExprDef`, the program would fail to compile with `Class not found : Binop`.
```
Simn: This "e.g." thing is also something for our future style-guide.

content/03-type-system.md
```diff
@@ -518,11 +518,11 @@ Haxe also allows the more natural `as` in place of `in`.
 
 ##### since Haxe 3.3.0
 
-This feature allows us to define default imports and usings that will be applied for all modules inside a directory, which is particularly handy for large code bases with a lot of helpers and static extensions because it reduces the number of imports.
+Using the specially named `import.hx` file (note the lowercase name), default imports and usings can be defined that will be applied for all modules inside a directory, which reduces the number of imports for large code bases with many helpers and static extensions.
 
-The special named `import.hx` file (note the lowercase name) can be placed in the directory where your code lies. The file can contain import and using statements only. The statements will be applied to all Haxe modules in this directory and its subdirectories.
+The `import.hx` file should be placed in the directory where your code resides. This file can contain import and using statements only. The statements will be applied to all Haxe modules in the directory and its subdirectories.
```
Simn: Isn't "should" still too weak here? I think this paragraph needs a rewrite.

content/03-type-system.md
```diff
@@ -562,19 +562,19 @@ Given an identifier `i`, the algorithm is as follows:
 8. If a static named `i` is explicitly imported, resolve to it and halt.
 9. If `i` starts with a lower-case character, go to 11.
 10. If a type named `i` is available, resolve to it and halt.
-11. If the expression is not in untyped mode, go to 14
+11. If the expression is not in untyped mode, go to 14.
 12. If `i` equals `__this__`, resolve to the `this` constant and halt.
 13. Generate a local variable named `i`, resolve to it and halt.
-14. Fail
+14. Fail.
```
Simn: Hmm, really? This isn't exactly a sentence...

content/03-type-system.md
```diff
@@ -221,7 +221,7 @@ Unification is the heart of the type system and contributes immensely to the rob
 
 > ##### Define: Unification
 >
-> Unification between two types A and B is a directional process which answers the question if A **can be assigned to** B. It may **mutate** either type if it is or has a [monomorph](types-monomorph).
+> Unification between two types A and B is a directional process which answers one question: if A **can be assigned to** B. It may **mutate** either type if it either is or has a [monomorph](types-monomorph).
```
Aurel300: `… question: whether A can be …`

content/03-type-system.md
```diff
@@ -294,15 +294,15 @@ Unification of types having or being a [monomorph](types-monomorph) is detailed
...
 
-The opposite applies as well: If a function declares a return type of `Void`, it cannot return `Dynamic` or any other type. However, this direction of unification is allowed when assigning function types:
+The opposite applies as well: if a function declares a return type of `Void`, it cannot return `Dynamic` or any other type. However, this direction of unification is allowed when assigning function types:
 
 ```haxe
 var func:Void->Void = function() return "foo";
 ```
 
-The right-hand function clearly is of type `Void->String`, yet we can assign it to the variable `func` of type `Void->Void`. This is because the compiler can safely assume that the return type is irrelevant, given that it could not be assigned to any non-`Void` type.
+The right-hand function is clearly of type `Void->String`, yet we can assign it to the variable `func` of type `Void->Void`. This is because the compiler can safely assume that the return type is irrelevant, given that it could not be assigned to any non-`Void` type.
```
Aurel300: Maybe we should use the new function type syntax here. `()->String` and `()->Void`

content/03-type-system.md
```diff
@@ -315,9 +315,9 @@ Given a set of multiple types, a **common base type** is a type which all types
 
 Although `Base` is not mentioned, the Haxe Compiler manages to infer it as the common type of `Child1` and `Child2`. The Haxe Compiler employs this kind of unification in the following situations:
 
-* array declarations
-* `if`/`else`
-* cases of a `switch`
+* Array declarations.
+* `if`/`else`.
+* Cases of a `switch`.
```
Aurel300: Are these the only cases? How about the ternary operator?

content/03-type-system.md
Aurel300: Do we collectively dislike e.g.?

content/03-type-system.md
```diff
@@ -475,41 +475,41 @@ In this example, we are actually importing a **module**, not just a specific typ
...
 
 [code asset](assets/ImportWildcard.hx)
 
-Using the wildcard import on `haxe.macro` allows accessing `Expr` which is a module in this package, but it does not allow accessing `ExprDef` which is a sub-type of the `Expr` module. This rule extends to static fields when a module is imported.
+Using the wildcard import on `haxe.macro` allows accessing `Expr`, which is a module in this package, but it does not allow accessing `ExprDef` which is a sub-type of the `Expr` module. This rule extends to static fields when a module is imported.
 
-When using wildcard imports on a package the compiler does not eagerly process all modules in that package. This means that these modules are never actually seen by the compiler unless used explicitly and are then not part of the generated output.
+When using wildcard imports on a package, the compiler does not eagerly process all modules in that package; modules that have not been used explicitly are not part of the generated output.
```
Aurel300: Extra space at the end of the line?

content/03-type-system.md
```diff
@@ -475,41 +475,41 @@ In this example, we are actually importing a **module**, not just a specific typ
...
+Using the wildcard import on `haxe.macro` allows accessing `Expr`, which is a module in this package, but it does not allow accessing `ExprDef` which is a sub-type of the `Expr` module. This rule extends to static fields when a module is imported.
 
-When using wildcard imports on a package the compiler does not eagerly process all modules in that package. This means that these modules are never actually seen by the compiler unless used explicitly and are then not part of the generated output.
+When using wildcard imports on a package, the compiler does not eagerly process all modules in that package; modules that have not been used explicitly are not part of the generated output. 
 
 ##### Import with alias
 
-If a type or static field is used a lot in an importing module it might help to alias it to a shorter name. This can also be used to disambiguate conflicting names by giving them a unique identifier.
+If a type or static field is used frequently in an imported module it might help to alias it to a shorter name. This can also be used to disambiguate conflicting names by giving them a unique identifier.
```
Aurel300: I think the original is hard to understand, but the revision has a different meaning. The "importing module" is the module that has `import` lines. If we use a particular type often in the actual code (say, `import my.pack.FooBarWithALongName;`), the original is suggesting aliasing it might be helpful (`import my.pack.FooBarWithALongName as Foo;`).

content/03-type-system.md
Aurel300: It is. Unlike the sentence fragments on lines 540 onward :)

content/03-type-system.md
Gama11: Though I think the convention for the new syntax is usually to have spaces around `->`?

RealyUniqueName: Is it? I don't like spaces around `->` :)

Aurel300: I don't like the spaces either, but it is arguably clearer and already in the manual, see https://haxe.org/manual/types-function.html

Gama11: It's also consistent with how arrow functions are usually formatted. Or do you not put spaces there either?

Aurel300: I put spaces when writing an arrow function, but not when writing a function type.

Gama11: Right, and the new function type syntax is very similar to / inspired by arrow functions, so it feels strange to me to use different styles for them.

content/03-type-system.md
ZanaDev: I don't mind adding anything else that should be added here, as long as you can give me a full list of what's needed. 

To be clear however, I'm not attempting to review or verify the accuracy of anything in the manual; that's for someone else to handle.

content/03-type-system.md
ZanaDev: Admittedly some changes might be a bit more of personal style leaking in, although I'm trying to stay consistent with the manual's existing approach to things. Nothing's explicitly wrong, but the change sounded better to me personally when read out loud.

content/03-type-system.md
ZanaDev: I'm unused to editing lists (not what I typically handle honestly) and made a style decision to preface every list with a colon before it begins and end every item with a period. I can revert these changes if you want.

content/03-type-system.md
ZanaDev: I have been removing it primarily because in most cases you're saving two words, either "for example" or "for instance," both of which are somewhat clearer.

If you'd prefer e.g. as a stylistic choice, I can revert similar changes back to it.

content/03-type-system.md
ZanaDev: Your change sounds appropriate. I should have noticed beforehand really with how much this paragraph annoyed me :)

content/03-type-system.md
Simn: I don't mind either way, but this strikes me as something where we should try to be consistent.

content/03-type-system.md
ZanaDev: It turns out that capitalization after colons can go either way, which I was unaware of (sorry). If desired I can shift it back to the way it was before, no big deal really.

I'll try and rephrase this section.

content/03-type-system.md
ZanaDev: On second thought I'll revert this. I was treating this change as if it was discussing two unique objects, instead of one with multiple states.

Simn:
Fyi, I have enabled the wiki for this repository (didn't realize it was enabled) and added https://github.com/HaxeFoundation/HaxeManual/wiki/Styleguide with no content. We can collect various things there as reference.

content/03-type-system.md
ZanaDev: >_>

content/03-type-system.md
ZanaDev: Sure, my bad for the information change.

content/03-type-system.md
ZanaDev: I intentionally didn't change this out of fear I might mess up the original intent. Easy fix.

ZanaDev:
I'll take some time to note down in an informal manner in the wiki why I'm making the decisions that I'm making so that they can be reviewed and discussed easier.

On a side note, I hope I've been approaching this editing process correctly. I do suddenly feel a bit strange about "imposing my will" so to speak; I know I kind of just spontaneously appeared out of nowhere. Regardless, I hope it's been a net positive so far.

content/03-type-system.md
Aurel300: I understand, I am not aiming to make this PR be the ultimate correction of this file. Seeing it just made me wonder, and the question is primarily for @Simn or @RealyUniqueName.

Aurel300:
@ZanaDev It is absolutely appreciated! Any contributions are welcome. AFAIK the manual in its current form never underwent a proper review process so it is still far from stylistically consistent.

content/03-type-system.md
ZanaDev: I'll leave this as is for now in this pull request and make a more global change (or not) once everyone's settled on how this should be handled. I think I'll just open an issue for discussion on general style stuff soon.

content/03-type-system.md
RealyUniqueName: Also for `try...catch...catch...`

Simn:
@ZanaDev Is this good to merge or did you want to make other changes?

Aurel300:
No response for a while, so let's merge it for now. We can get the other comments implemented in a separate PR.

@ZanaDev Thank you for the contribution!

PR #450, opened 2020-01-28
Update 10-std.md

aszasz:


Aurel300:
Thanks, probably a leftover from LaTeX.

PR #451, opened 2020-02-04
Add documentation for --run

haxiomic:
Copied doc from `haxe --help`


Aurel300:
This is fine. I think the docs for these can be improved, but that's not specific to `--run`. Also in the future I think it would be nice if these were also JSON-ified like metas and defines.

PR #454, opened 2020-03-20
Document @:ifFeature

RealyUniqueName:


RealyUniqueName:
@Simn please check if that's correct.

Simn:
Well, "not removed by DCE" is a necessary condition but not sufficient. The field also has to be seen by the compiler in the first place.

PR #455, opened 2020-03-24
Document `haxe.Exception`

RealyUniqueName:
Also moved "throw" section right before "try/catch"

### DIFF
--- a/content/05-expression.md
+++ b/content/05-expression.md
@@ -795,7 +795,7 @@ A value which is thrown like this can be caught by [`catch` blocks](expression-t
 ##### since Haxe 4.1.0
 
 It's highly recommended to not throw arbitrary values and instead throw instances of `haxe.Exception`.
-In fact if `value` is not an instance of `haxe.Exception`, then `throw value` is compiled as `throw haxe.Exception.thrown(value)`, which wraps `value` into an instance of `haxe.Exception`.
+In fact, if `value` is not an instance of `haxe.Exception`, then `throw value` is compiled as `throw haxe.Exception.thrown(value)`, which wraps `value` into an instance of `haxe.Exception`.
 
 However native target exceptions are thrown as-is. For example an instance of `cs.system.Exception` or `php.Exception` won't get automatically wrapped upon throwing.
 
@@ -831,13 +831,13 @@ This process has many similarities to the compile-time [unification](type-system
 
 ##### Haxe 3.* and Haxe 4.0
 
-Prior to Haxe 4.1.0 to catch an exception of any type it is required to use `Dynamic` catch type.
+Prior to Haxe 4.1.0 the only way to catch all exceptions is by using `Dynamic` as the catch type.
 To get a string representation of the exception `Std.string(e)` could be used. 
 ```haxe
 try {
   doSomething();
 } catch(e:Dynamic) {
-  //All exceptions will be caught here
+  // All exceptions will be caught here
   trace(Std.string(e));
 }
 ```
@@ -858,7 +858,7 @@ try {
 
 ##### Haxe 3.* and Haxe 4.0
 
-Exception call stack is available via `haxe.CallStack.exceptionStack()` inside of a `catch` block:
+The exception call stack is available via `haxe.CallStack.exceptionStack()` inside of a `catch` block:
 ```haxe
 try {
   doSomething();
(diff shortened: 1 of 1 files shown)


content/05-expression.md
```diff
@@ -781,6 +781,26 @@ switch (0) {
...
+```
+
+A value which is thrown like this can be caught by [`catch` blocks](expression-try-catch). If no such block catches it, the behavior is target-dependent.
+
+##### since Haxe 4.1.0
+
+It's highly recommended to not throw arbitrary values and instead throw instances of `haxe.Exception`.
+In fact if `value` is not an instance of `haxe.Exception`, then `throw value` is compiled as `throw haxe.Exception.thrown(value)`, which wraps `value` into an instance of `haxe.Exception`.
+
+However native target exceptions are thrown as-is. For example an instance of `cs.system.Exception` or `php.Exception` won't get automatically wrapped upon throwing.
```
Simn: Something to clarify (maybe in the section below) is that these native exception classes can still be caught with `catch (e:haxe.Exception`).

content/05-expression.md
```diff
@@ -798,14 +818,184 @@ If during runtime the evaluation of `try-expression` causes a [`throw`](expressi
...
+```
+Main.hx:13: 
+Called from Main.doSomething (Main.hx line 11 column 15)
+Called from Main.main (Main.hx line 5 column 5)
+Main.hx:17: 
+Called from Main.doSomething (Main.hx line 11 column 15)
+Called from Main.main (Main.hx line 5 column 5)
+```
+
+If `catch(e:haxe.Exception)` catches a native exception, then rethrowing `e` throws that native exception instead of a `haxe.Exception` instance.
```
Simn: Hmm, now I'm a bit confused because this would imply that you _can't_ catch any exception with `catch (e:haxe.Exception)`.

RealyUniqueName: I don't see why? Maybe my English skills fail me )

Simn: It says that a native exception is thrown _instead_ of a `haxe.Exception` instance. This tells me that the native exception is not a `haxe.Exception` instance, and it follows that it can't be caught as such.

RealyUniqueName: Would adding "but it still could be caught with `catch(e:haxe.Exception)`" help?

content/05-expression.md
```diff
@@ -798,14 +818,184 @@ If during runtime the evaluation of `try-expression` causes a [`throw`](expressi
...
+      heavyCalculation();
+    } catch(e:Exception) {
+      throw new MyLibException(e.message, e);
+    }
+  }
+
+  static function heavyCalculation() {}
+}
+```
+Now library users don't have to bother about specific arithmetic exceptions. All they need to do is handle `MyLibException`
```
Gama11: I'd phrase this as "don't have to _worry_ about" instead. Also, missing a dot at the end. :)

content/05-expression.md
```diff
@@ -798,14 +818,184 @@ If during runtime the evaluation of `try-expression` causes a [`throw`](expressi
...
+    cleanup();
+    throw new haxe.Exception('Failed to do something', e);
+  }
+} catch(e:haxe.Exception) {
+  trace(e.message); // "Failed to do something"
+  trace(e.previous.message); // "Terrible error"
+}
+```
+
+Another use-case is creating a library, which does not expose internal exceptions to public API, but still provides information about exceptions reasons:
```
Gama11: I think "_as_ public API" sounds better.

Gama11:
I didn't care about the exception refactor all that much / didn't follow it too much, but this really shows how much nicer everything is now. 👍

content/05-expression.md
```diff
@@ -798,14 +818,185 @@ If during runtime the evaluation of `try-expression` causes a [`throw`](expressi
...
+#### Wildcard catch
+
+##### Haxe 3.* and Haxe 4.0
+
+Prior to Haxe 4.1.0 the only way to catch all exceptions is by using `Dynamic` as the catch type.
+To get a string representation of the exception `Std.string(e)` could be used. 
+```haxe
+try {
+  doSomething();
+} catch(e:Dynamic) {
```
markknol: Shouldn't we encourage to use `Any` here?

RealyUniqueName: Probably yes. But I've never seen any Haxe code catching `Any` and it's too late to encourage it anyway because now we should encourage `haxe.Exception` instead :)

markknol: I always use `catch(error:Any)`, if I don't know what's going on.

RealyUniqueName: Alright! Alright! I'll do it! :)

PR #461, opened 2020-05-14
Tail Recursion Elimination

RealyUniqueName:


content/08-cr-features.md
```diff
@@ -585,6 +585,54 @@ Whether a loop is unrolled or not depends on the loop unrolling cost, which is d
...
+```haxe
+  static function getRoot(node:Node):Node {
+    var root = node;
+    while(root.parent != null) {
+      root = root.parent;
+    }
+    return root;
+  }
+```
+Its recursive counterpart may look cleaner, but has worse performance because of additional function calls:
```
Simn, line 606: I guess it's better to be a bit vague here and say "might have" instead of "has".

content/08-cr-features.md
```diff
@@ -585,6 +585,54 @@ Whether a loop is unrolled or not depends on the loop unrolling cost, which is d
...
+```haxe
+  static function getRoot(node:Node):Node {
+    if(node.parent == null) {
+      return node;
+    }
+    return getRoot(node.parent);
+  }
+```
+
+Luckily Haxe compiler since version 4.1.0 can perform automatic recursion-to-loop transformation for functions which end with a recursive call.
```
Simn, line 616: "the Haxe compiler"

Aurel300: Also "luckily" seems not to fit the general tone.

content/08-cr-features.md
```diff
@@ -585,6 +585,54 @@ Whether a loop is unrolled or not depends on the loop unrolling cost, which is d
...
+    if (node.parent == null) return node;
+    node = node.parent;
+  }
+}
+```
+The recursive call is automatically replaced with a loop on compilation.
+
+A function must satisfy a number of requirements to be eligible for TRE optimisation:
+1. The last operation of the function is a recursive call. Even if that call is inside of an `if` or `else` expressions.
+2. The function does not have [`dynamic` accessor](/manual/class-field-dynamic.html).
```
Aurel300, line 631: "a dynamic accessor"

content/08-cr-features.md
```diff
@@ -585,6 +585,54 @@ Whether a loop is unrolled or not depends on the loop unrolling cost, which is d
...
+  static function getRoot(node:Node):Node {
+    if(node.parent == null) {
+      return node;
+    }
+    return getRoot(node.parent);
+  }
+```
+
+Luckily Haxe compiler since version 4.1.0 can perform automatic recursion-to-loop transformation for functions which end with a recursive call.
+This optimisation is automatically enabled with the [static analyzer](/manual/cr-static-analyzer.html) (by adding `-D analyzer-optimize` to compilation arguments).
```
Aurel300, line 617: As much as I like BrEng, I think we are consistent with American spelling in the manual. "optimization".

PR #465, opened 2020-06-16
Update 04-class-field.md

RichardBray:


content/04-class-field.md
```diff
@@ -393,6 +393,7 @@ The `extern` keyword causes the compiler to not generate the field in the output
 The `final` keyword can be used on class fields with the following effects:
```
nadako, line 393: this `on class fields` doesn't sound accurate now (even before your change, but still)

RichardBray: Ah good point, maybe I should move my line to outside of line 393, as in above it.

PR #473, opened 2020-11-25
HashLink -> lib requirement in compile.hxml added

LuminousLizard:
"--lib hlsdl" or "--lib hldx" in compile.hxml are required, otherwise the execution by Hashlink will not start. Information is in the Readme.md but not in the doc.

### DIFF
--- a/content/12-target-details.md
+++ b/content/12-target-details.md
@@ -2306,6 +2306,7 @@ Another possibility is to create and run (double-click) a file called `compile.h
 
 ```hxml
 --hl hello.hl
+--lib hlsdl # SDL support (Linux, Windows, MacOS), use '--lib hldx' for DirectX support (only Windows)
 --main Main
 ```
 


Aurel300:
I don't think this is true (unless changed recently, which I doubt). You can write a console-only application, like the "Hello World" example is, without needing a graphics library. Were you trying to run a heaps example maybe?

LuminousLizard:
@Aurel300 yeah thats true .. its a heap example. I found it now on the heaps docs. This can be discarded.
But it is "nice" that the use of HashLink commands is not explained there. https://heaps.io/documentation/hashlink.html

Aurel300:
I suppose that part of the documentation should be covered by the "Hello Hashlink" section. But you could open a patch to the [Heaps documentation](https://github.com/HeapsIO/heaps-doc) repository.

PR #476, opened 2021-02-01
specify that --cs does in fact generate binaries

mininmobile:
specify that `--cs` does in fact generate binaries just like `--cpp`


Simn:
This is wonky grammar, but I realize that it was just copied from the wonky grammar in the C++ section, so whatever. Thanks!

PR #487, opened 2021-08-13
Add example of Array comprehesion expression

jmdejong:
This adds an example of how to use expressions in a for loop array comprehension.

This is something found in other programming languages (and yet others use `map` for this) so for programmers who know this it will be valuable to see a minimal example of how it's done in haxe.

### DIFF
--- a/assets/ArrayComprehension.hx
+++ b/assets/ArrayComprehension.hx
@@ -6,5 +6,8 @@ class Main {
     var i = 0;
     var b = [while (i < 10) i++];
     trace(b); // [0,1,2,3,4,5,6,7,8,9]
+
+    var squares = [for (i in 0...10) i*i];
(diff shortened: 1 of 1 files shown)


Simn:
While this might be a worthwhile addition, it's not a good place for it. The example in question appears in the context of https://haxe.org/manual/lf-array-comprehension.html, so randomly adding two lines of code there doesn't work so well.

jmdejong:
I added these lines because it appears in that article.

That article does not have a good example of how the expression can be used to perform the same functionality as a map function.

The first example only puts the same values that are being iterated over in the array.
There is a more complex example on that page, but that also introduces multiple for loops being used in the same comprehension.
I thought it would be useful to have an example that's somewhat in between in complexity.

PR #489, opened 2021-08-18
Fix the email regex example

jmdejong:
The dot was not escaped, which would allow an email like "bob@example/com" to be matched.

The example assumed that domain names are 2 or 3 characters long.
This is not true anymore (see https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains).

There are still a lot of ways to be pedantic about email adresses because technically a lot more addresses are valid, but this matches most addresses in practice.


Simn:
Thanks!

PR #491, opened 2021-08-28
Improve definition of `private` in Haxe

SenkJu:
The documenation says that Haxe does not support the keyword `protected` known from many other object-oriented languages, but its `private` behaves analogously. This is only partially true, however, because Java, for example, allows access to methods/fields even from non-inheriting classes in the same package when declared as `protected`. Haxe does not.

Oracle Docs
[See docs](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html)

I have updated the description of Haxe's `private` behavior to better convey what it does.

### DIFF
--- a/content/04-class-field.md
+++ b/content/04-class-field.md
@@ -272,7 +272,7 @@ Omitting the visibility modifier usually defaults the visibility to `private`, b
 
 > ##### Trivia: Protected
 >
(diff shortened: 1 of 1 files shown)


Simn:
I don't like that this paragraph now states that Haxe doesn't support the `protected` _behavior_. The whole point here is that it largely does. Let's stick with "keyword" here. Otherwise I'm fine with these changes.

Simn:
Thanks!

PR #493, opened 2021-09-21
Add documentation and examples for abstract classes

### LINKED ISSUE #492: Provide documentation for abstract classes.
Abstract classes are undocumented in the manual

- [ ] Create documentation and example use of abstract classes.

ubuntom:
Fixes #492

### DIFF
--- a/content/02-types.md
+++ b/content/02-types.md
@@ -308,6 +308,123 @@ Like classes, interfaces can be marked with the `final` keyword, preventing them
 
 
 
+<!--label:types-abstract-class-->
+#### Abstract Class
+
+##### since Haxe 4.2.0
+
+Abstract classes (not to be confused with [Abstract](type-system-type-inference)) are classes with partial implementation. As such, an abstract class cannot be instantiated but must first be extended, with the child class providing implementation of any abstract methods. Abstract classes support all language features that concrete classes support, thus any class can be declared abstract.
+
+Concrete classes inheriting from abstract classes must be declared without the `abstract` keyword, and all abstract methods in the parent classes must be implemented.
+
+Abstract class methods behave similarly to interfaces; implementations of the methods do not use the `override` keyword.
+
+```haxe
+abstract class Vehicle {
+	var speed:Float = 0;
+
+	abstract public function getWheels():Int;
+
+	public function new() {}
+}
+
+class Car extends Vehicle {
+	public function getWheels() {
+		return 4;
+	}
+
+	public function accelerate() {
+		speed += 1;
+	}
+}
+
+class Bike extends Vehicle {
+	public function getWheels() {
+		return 2;
+	}
+
+	public function accelerate() {
+		speed += 2;
+	}
+}
+```
+
+An abstract class can be used to provide a partial implementation of an interface, leaving the implementation of other methods up to the child classes.
+
+```haxe
+interface Vehicle {
+	public function getFuelType():String;
+	public function getWheels():Int;
+}
+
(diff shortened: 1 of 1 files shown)


content/02-types.md
```diff
@@ -308,6 +308,123 @@ Like classes, interfaces can be marked with the `final` keyword, preventing them
 
 
 
+<!--label:types-abstract-class-->
+#### Abstract Class
+
+##### since Haxe 4.2.0
+
+Abstract classes (not to be confused with [Abstract](type-system-type-inference)) are classes with partial implementation. As such, an abstract class cannot be instantiated but must first be extended, with the child class providing implementation of any abstract methods. Abstract classes support all language features that concrete classes support, thus any class can be declared abstract.
```
Simn, line 316: We should say "with the child class providing implementation of any abstract methods or being declared abstract itself".

content/02-types.md
```diff
@@ -308,6 +308,123 @@ Like classes, interfaces can be marked with the `final` keyword, preventing them
...
+		// var spaceship = new Spaceship();  // Fail
+		var rocket = new Rocket(); // Succeed
+	}
+}
+
+```
+
+> ##### Trivia: Constructor
+>
+> Even though they can't be instantiated, abstract classes can still have a constuctor.
```
Simn, line 424: I wouldn't trivia this part, and we should also mention the reason: Child classes can call these constructors via `super()`.

And given that everything else here has these nice code examples, I suppose this one should get one as well.

content/02-types.md
```diff
@@ -308,6 +308,123 @@ Like classes, interfaces can be marked with the `final` keyword, preventing them
...
 
 
+<!--label:types-abstract-class-->
+#### Abstract Class
+
+##### since Haxe 4.2.0
+
+Abstract classes (not to be confused with [Abstract](type-system-type-inference)) are classes with partial implementation. As such, an abstract class cannot be instantiated but must first be extended, with the child class providing implementation of any abstract methods. Abstract classes support all language features that concrete classes support, thus any class can be declared abstract.
+
+Concrete classes inheriting from abstract classes must be declared without the `abstract` keyword, and all abstract methods in the parent classes must be implemented.
```
Simn, line 318: Maybe we should properly define the "concrete" here, it comes as a bit of a surprise.

Simn:
Thanks a lot! I've left a few remarks.

content/02-types.md
```diff
@@ -377,53 +377,62 @@ As with other inheritance relationships, child classes can be assigned to a type
...
+    var rocket = new Rocket(); // Succeed
+  }
 }
-
 ```
 
-> ##### Trivia: Constructor
->
-> Even though they can't be instantiated, abstract classes can still have a constuctor. 
+Even though they can't be instantiated, abstract classes can still have a constructor that the child class can call with `super()`. Definition of this constructor is optional.
```
Simn, line 420: > Definition of this constructor is optional.

Hmm, what exactly do you mean here?

ubuntom: That's meant to say that the abstract class doesn't _need_ the constructor. 
Perhaps the following code should be the example:
`abstract class Parent {
}

class Child extends Parent {
	public function new() {
    trace("Child created!");
  }
}
`

Simn: Right, I suppose that's already implied by the "can" in the previous sentence. I think this is actually clearer without the additional line because it doesn't really clarify anything and might just confuse people (like me just now, haha).

ubuntom: OK, I'll drop that last sentence

PR #495, opened 2021-09-29
Array initialization example

Pidroh:
Added an example to array initialization since I felt like the manual was unclear


Simn:
I'm fine with that, but let's please lose the ` : Array` type-hint because that's not very haxe-idomatic.

Pidroh:
@Simn 
This is what you meant? I'm quite a beginner at Haxe.

Simn:
Yes, but please do it for both lines!

Pidroh:
Oh, that was dumb. Done!

Simn:
Thanks!

The fact that `b` appears before `a` is pretty strange, but I feel rebellious today so I'll merge it anyway.

PR #496, opened 2021-10-08
Added overload and rest arguments section

MrcSnm:
Added function overload 
Added function rest arguments


Geokureli:
Everyone in the haxe discords is losing their shit because no one knew these were possible. Why was this not documented?

PR #506, opened 2022-07-11
Change instructions regarding haxelib

ConfidantCommunications:
Since the documentation already has the awkwardness of acknowledging an out of date haxelib plus a workaround, I thought this would be a more straightforward way of describing installation.


Simn:
Can we just omit this whole "out of date" part? All it does is make us look sloppy...

ConfidantCommunications:
I'm good with that. I can submit an adjustment…

ConfidantCommunications:
Okay I updated the file (hopefully did it properly).

PR #508, opened 2022-08-13
Updated string literal documentation

TheTophatDemon:
Included paragraph explaining that string literals can occupy multiple lines, rather than there being a special language construct for doing so.

### DIFF
--- a/content/10-std.md
+++ b/content/10-std.md
@@ -40,10 +40,24 @@ String literals may occupy multiple lines; In which case, each line of the strin
 
 ```haxe
 var str = "Line one
-	Line two
-	Line three"; //Lines two and three have \t in them
+Line two
+Line three";
 trace(str);
 ```
+
+Note that indentation will also be included in the string, such as with the example below.
+
+```haxe
+class X {
+  function foo() {
+    var str = "a
+    b
+    c";
+  }
+}
+```
+
+The `str` variable will have a '\t' tab character before 'b' and 'c'.
 
 ##### Escape sequences
 
--- a/content/10-std.md
+++ b/content/10-std.md
@@ -36,6 +36,15 @@ trace(a == b); // true
 
 The only difference between the two forms is that single-quoted literals allow [string interpolation](lf-string-interpolation).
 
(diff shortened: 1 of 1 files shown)


content/10-std.md
```diff
@@ -36,6 +36,15 @@ trace(a == b); // true
 
 The only difference between the two forms is that single-quoted literals allow [string interpolation](lf-string-interpolation).
 
+String literals may occupy multiple lines; In which case, each line of the string will be punctuated with a '\n' newline character (See the escape sequences chart below).
+
+```haxe
+var str = "Line one
+	Line two
+	Line three"; //Lines two and three have \t in them
```
Aurel300, line 44: The point of this addition is to demonstrate new lines. I think it's better to show this without the tabs, it might be confusing. You could have one example where the second line onwards is not indented, then show that if you indent the string as you would the code in a function, e.g.

```
class X {
  function foo() {
    var str = "a
    b
    c";
  }
}
```

Then the indents become part of the string. Does that make sense?

TheTophatDemon: That makes sense; I've implemented that change in a new commit. Should I merge it with the previous one?

content/10-std.md
```diff
@@ -40,11 +40,25 @@ String literals may occupy multiple lines; In which case, each line of the strin
...
+class X {
+  function foo() {
+    var str = "a
+    b
+    c";
+  }
+}
+```
+
+The `str` variable will have a '\t' tab character before 'b' and 'c'.
```
Aurel300, line 60: Nit: it now has 4 spaces, not a tab.

content/10-std.md
Aurel300: PRs here are squashed, so it does not matter whether it is a separate commit.

content/10-std.md
```diff
@@ -36,6 +36,29 @@ trace(a == b); // true
...
```
Aurel300, line 60:
```suggestion
The `str` variable will have four spaces before 'b' and 'c'.
```

Aurel300:
Thanks!

PR #513, opened 2022-09-28
Document constructor inlining limitations.

player-03:
[Here's some code demonstrating the limitations](https://try.haxe.org/#b4E6abfa).

```haxe
class InlineConstructorTest {
  static function main() {
    //Can't be inlined because `object1` is reassigned later.
    var object1:SimpleObject = new SimpleObject("one");
    object1 = new SimpleObject("uno");
    
    //Can be inlined.
    var object2:SimpleObject = new SimpleObject("two");
    object2.name = "dos";
    
    //Can't be inlined because `object3` is passed to a function.
    var object3:SimpleObject = new SimpleObject("three");
    if(Reflect.fields(object3).length > 0) {
      object3.name = "tres";
    }
    
    trace(object1.name);
    trace(object2.name);
    trace(object3.name);
    printObject4();
  }
  
  static inline function printObject4():SimpleObject {
    //Can't be inlined because `object4` is returned.
    var object4:SimpleObject = new SimpleObject("cuatro");
    trace(object4.name);
    return object4;
  }
}

class SimpleObject {
  public var name:String;
  
  public inline function new(name:String) {
    this.name = name;
  }
}
```

### DIFF
--- a/content/06-lf.md
+++ b/content/06-lf.md
@@ -939,11 +939,11 @@ This is also true for access granted to parent classes, in which case it extends
 
 ##### since Haxe 3.1.0
 
-If a constructor is declared to be [inline](class-field-inline), the compiler may try to optimize it away in certain situations. There are several requirements for this to work:
+If a constructor is declared to be [inline](class-field-inline), the compiler may try to optimize it away in certain situations. For this to work, the compiler must be able to verify that the object can be replaced by a set of local variables.
 
-* The constructor must only contain assignments to its instance fields.
-* The newly-constructed instance must only be accessible from the enclosing function.
-* If the new instance is assigned to a variable, that variable must be deconstructible, meaning among other things, it may not be reassigned.
+* The newly-constructed instance must only be accessible from the local function.
+* All called instance functions must be inlined.
+* All other references to the instance must only read or write its variables.
 
 The following example demonstrates constructor inlining:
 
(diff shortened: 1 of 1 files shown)


Simn:
This might be a bit too specific because it's really about any use of the local variable in a way that doesn't allow deconstruction. Not sure how to express that succinctly though.

player-03:
> We can also add that `extern inline function` declaration disables deinlining, and compiler starts showing errors in cases, when inline is not possible.

This is also true if you write `inline new MyClass()`, [as mentioned at the bottom of the `inline` keyword page](https://haxe.org/manual/expression-inline.html). Probably should be mentioned here too, but I might make a different pull request for that.

Simn:
I think that last line raises more questions than it answers... Introducing the term "deconstructible" doesn't help much.

Maybe instead of talking about all the things we can't do, it's easier to talk about what we _can_ do with the local variable, which is accessing its fields.

player-03:
Better?

> If a constructor is declared to be [inline](https://github.com/HaxeManual/HaxeManual/blob/patch-1/content/class-field-inline), the compiler may try to optimize it away in certain situations. For this to work, the compiler must be able to verify that the object can be replaced by a set of local variables.
> 
> - The newly-constructed instance must only be accessible from the local function.
> - All called instance functions must be inlined.
> - All other references to the instance must only read or write its variables.

Simn:
Seems good now. Thank you for your patience!

player-03:
> Thank you for your patience!

This only took three days and three revisions. From my perspective that's really fast.

PR #514, opened 2022-09-30
Add `@:keep` to sample function

### LINKED ISSUE #475: haxe.Template  sample fails with DCE
Hej !

```haxe
class Main {
  static function main() {
    new Main();
  }

  public function new() {
    var user = {name: "Mark", distance: 3500};
    var sample = "The results: $$display(::user::,::time::)";
    var template = new haxe.Template(sample);
    var output = template.execute({user: user, time: 15}, this);
    trace(output);
  }

  function display(resolve:String->Dynamic, user:User, time:Int) {
    return user.name + " ran " + (user.distance / 1000) + " kilometers in " +
      time + " minutes";
  }
}

typedef User = {name:String, distance:Int}
```

This sample https://haxe.org/manual/std-template.html fails in PHP 7 with Haxe Compiler 4.2.0-rc.1+5145ddb
Error : 
```haxe
Fatal error:  Uncaught Macro call display(<function>,{ name : Mark, distance : 3500 },15) failed (ErrorException: call_user_func_array() expects parameter 1 to be a valid callback, no array or string given in /lib/Reflect.php:37
Stack trace:
#0 /lib/Reflect.php(37): php\Boot::php\{closure}(2, 'call_user_func_...', '/www/o...', 37, Array)
HaxeFoundation/haxe#1 /lib/haxe/Template.php(916): Reflect::callMethod(Object(Server), NULL, Object(Array_hx))
HaxeFoundation/haxe#2 /lib/haxe/Template.php(826): haxe\Template-&gt;run(Object(haxe\_Template\TemplateExpr))
HaxeFoundation/haxe#3 /lib/haxe/Template.php(127): haxe\Template-&gt;run(Object(haxe\_Template\TemplateExpr))
HaxeFoundation/haxe#4 /lib/Server.php(163): haxe\Template-&gt;execute(Object(_HxAnon_Server4), Object(Server))
HaxeFoundation/haxe#5 /lib/Server.php(42): Server-&gt;__construct()
HaxeFoundation/haxe#6 /www/index.php(14): Server::main()
HaxeFoundation/haxe#7 {main})
  thrown in /lib/haxe/Exception.php on line 96
```

player-03:
Otherwise, DCE would clean it up and break the sample. Granted, you wouldn't typically run a sample with DCE, but if users were to copy the code into their own project, they could run into the issue there.

Resolves #475.

### DIFF
--- a/assets/TemplateMacros.hx
+++ b/assets/TemplateMacros.hx
@@ -11,7 +11,7 @@ class Main {
     trace(output);
   }
 
-  function display(resolve:String->Dynamic, user:User, time:Int) {
+  @:keep function display(resolve:String->Dynamic, user:User, time:Int) {
     return user.name + " ran " + (user.distance / 1000) + " kilometers in " +
       time + " minutes";
   }
--- a/content/10-std.md
+++ b/content/10-std.md
@@ -615,7 +615,8 @@ The console will trace `The users are Mark(30) John(45)`.
 ##### Template macros
 To call custom functions while parts of the template are being rendered, provide a `macros` object to the argument of [Template.execute](https://api.haxe.org/haxe/Template.html#execute). The key will act as the template variable name, the value refers to a callback function that should return a `String`. The first argument of this macro function is always a `resolve()` method, followed by the given arguments. The resolve function can be called to retrieve values from the template context. If `macros` has no such field, the result is unspecified.
 
-The following example passes itself as macro function context and executes `display` from the template.
+The following example passes itself as macro function context and executes `display` from the template. Because `display` isn't called anywhere else, `@:keep` is used to prevent DCE from removing it.
+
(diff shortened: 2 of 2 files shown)


Simn:
In that case we should probably also mention why we have `@:keep` there.

player-03:
How does that wording look?

Simn:
"anywhere else" sounds somewhat imprecise, but I don't want to make this too complicated here. Thank you!

player-03:
Possible alternatives:

- Because the template uses reflection to call `display`, `@:keep` is used to prevent DCE from removing it.
- Because `display` is only called using reflection, `@:keep` is used to prevent DCE from removing it.
- `@:keep` is used to prevent DCE from removing `display`.
- The `@:keep` metadata lets DCE know that `display` isn't unused.

PR #520, opened 2023-01-16
Fix the error that Page Not Found

Frederisk:
These pages have been removed in the latest haxe api docs. Use an older version of the documentation to avoid 404 Page Not Found errors.


Aurel300:
Thanks!

PR #523, opened 2023-04-02
Update lua dependencies

tobil4sk:
Regex library is handled separately in #522

### DIFF
--- a/content/12-target-details.md
+++ b/content/12-target-details.md
@@ -2180,7 +2180,6 @@ With LuaRocks, install the following packages:
 
 ```sh
 luarocks install lrexlib-pcre2
-luarocks install environ
 luarocks install luasocket
 luarocks install luasec
 luarocks install luv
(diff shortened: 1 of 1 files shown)


Simn:
This conflicts after merging the other PR.

And is this really saying that this should install this library on 5.1 and 5.4, but not on 5.3 (which it used to say).

tobil4sk:
Yes, it was incorrect before. For 5.2 and 5.3, the bit32 library is a built-in part of std so it does not need to be installed manually. 5.1 didn't have this library yet, so someone created a backport and uploaded it to luarocks, which is why you need to run `luarocks install bit32` on 5.1.

Then, 5.4 removed this library in favour of native bitwise operators, which is why the bit32 library (originally intended just as a backport for 5.1, but now usable on anything < 5.5) should be installed on 5.4 as well.

Our unit tests don't install bit32 from luaroks on 5.3 and they run fine: https://github.com/HaxeFoundation/haxe/blob/dd0d6a64e90f3b93b8a1be966184f51bc6d35903/tests/runci/targets/Lua.hx#L80-L83.

RblSb:
Our tests also depend on specific old versions of luasocket / luv / luautf8, for example you cannot install old enough luautf8 on lua 5.4 to make `length` work on broken unicode strings, instead of returning `nil` (can affect http requests and other stuff). I'm not sure about best solution here.

PR #526, opened 2023-04-15
Update for 4.3

### LINKED ISSUE #524: Haxe 4.3 documentation TODO
* [x] defaults for type parameters (HaxeFoundation/haxe#10518)
* [x] support abstract keyword to reference the abstract (HaxeFoundation/haxe#10513)
* [x] support static var at expression-level (HaxeFoundation/haxe#10555)
* [x] support ?. safe navigation operator (HaxeFoundation/haxe#10561)
* [x] added ?? null coalescing operator (HaxeFoundation/haxe#10428)
* [ ] add -w compiler option to configure warnings (HaxeFoundation/haxe#10612)
* [x] support custom metadata and defines (HaxeFoundation/haxe#10858)

Aurel300:
Closes #524 (except for the `-w` comment).

- new section in Types: [Access to Underlying Data](https://github.com/HaxeFoundation/HaxeManual/blob/feature/4.3/content/02-types.md#access-to-underlying-data) (about the `abstract` keyword)
- expanded section on [operator overloading](https://github.com/HaxeFoundation/HaxeManual/blob/feature/4.3/content/02-types.md#since-haxe-430-1) to explain `@:op(a())`
- new section in Type System: [Default type parameters](https://github.com/HaxeFoundation/HaxeManual/blob/feature/4.3/content/03-type-system.md#defaults)
- new section in Expressions: [Local static variables](https://github.com/HaxeFoundation/HaxeManual/blob/feature/4.3/content/05-expression.md#local-static-variables)
  - (also added a note to static class fields linking to this)
- expanded section on [field access](https://github.com/HaxeFoundation/HaxeManual/blob/feature/4.3/content/05-expression.md#since-haxe-430) with null-safe field access
- new section in Expressions: [Null coalescing](https://github.com/HaxeFoundation/HaxeManual/blob/feature/4.3/content/05-expression.md#null-coalescing)
- new section in Language Features: [Custom defines documentation](https://github.com/HaxeFoundation/HaxeManual/blob/feature/4.3/content/06-lf.md#custom-defines-documentation)
- new section in Language Features: [Custom metadata documentation](https://github.com/HaxeFoundation/HaxeManual/blob/feature/4.3/content/06-lf.md#custom-metadata-documentation)
- new section in Compiler Usage: [Error Reporting](https://github.com/HaxeFoundation/HaxeManual/blob/feature/4.3/content/07-compiler-usage.md#error-reporting)
- expanded section on [expression reification](https://github.com/HaxeFoundation/HaxeManual/blob/feature/4.3/content/09-macro.md#since-haxe-430) to explain that `$` escapes are not allowed outside of `macro` expressions

Left to do:

- [ ] `-w` documentation (@simn wanted to take this?)

### DIFF
--- a/content/02-types.md
+++ b/content/02-types.md
@@ -20,7 +20,7 @@ We will explore type inference in detail later in [Type Inference](type-system-t
 
 The Haxe type system knows seven type groups:
 
- * **Class instance**: an object of a given class or interface 
+* **Class instance**: an object of a given class or interface 
 * **Enum instance**: a value of a Haxe enumeration 
 * **Structure**: an anonymous structure, i.e. a collection of named fields 
 * **Function**: a compound type of several arguments and one return 
@@ -870,6 +870,27 @@ One problem may be apparent - what happens if a member function is not declared 
 >
 > Before the advent of abstract types, all basic types were implemented as extern classes or enums. While this nicely took care of some aspects such as `Int` being a "child class" of `Float`, it caused issues elsewhere. For instance, with `Float` being an extern class, it would unify with the empty structure `{}`, making it impossible to constrain a type to accept only real objects.
 
+<!--label:types-abstract-this-->
+#### Access to Underlying Data
+
+As shown in the example from the previous section, `this` in abstract methods refers to the underlying data. For `AbstractInt` methods, `this` is therefore a variable of type `Int`.
+
+[code asset](assets/MyAbstract.hx#L1-L5)
+
+##### since Haxe 4.3.0
+
+The `abstract` keyword can be used in abstract methods to refer to the current instance *as an abstract type*, rather than referring to the underlying data with the `this` keyword.
+
+This can be useful when an abstract method needs to call other methods which accept an argument of the abstract type. For example, suppose class `Main` defines a method `takeAbstractInt`:
+
+[code asset](assets/MyAbstract2.hx#L11-L13)
+
+To call `takeAbstractInt` from within an `AbstractInt` method, we must use the `abstract` keyword:
+
+[code asset](assets/MyAbstract2.hx#L5-L7)
+
+If we instead wrote `Main.takeAbstractInt(this)`, this would be a type error, because `takeAbstractInt` does not accept an argument of type `Int`.
+
 <!--label:types-abstract-implicit-casts-->
 #### Implicit Casts
 
@@ -969,7 +990,11 @@ The method body of an `@:op` function can be omitted, but only if the underlying
 
 [code asset](assets/AbstractExposeTypeOperations.hx)
 
+##### since Haxe 4.3.0
 
+The `@:op(a())` syntax can be used to overload function calls on abstracts. The metadata is attached to a function, and the signature of that function determines the signature of the call to the abstract. Multiple functions with different signatures can be annotated this way to support overloading:
+
+[code asset](assets/AbstractCallOverload.hx)
 
 <!--label:types-abstract-array-access-->
 #### Array Access
--- a/content/03-type-system.md
+++ b/content/03-type-system.md
@@ -95,6 +95,14 @@ When constraining to a single type, the parentheses can be omitted:
 
 One of the breaking changes between versions 3 and 4 is the multiple type constraint syntax. As the first example above shows, in Haxe 4 the constraints are separated by an `&` symbol instead of a comma. This is similar to the new [structure extension](types-structure-extensions) syntax.
 
+<!--label:type-system-type-parameter-defaults-->
+#### Defaults
+
+Since Haxe 4.3 it is possible to specify a default type for each type parameter of a type. When using such a type in type annotations, type parameters with a default can be omitted:
+
+[code asset](assets/TypeParameterDefault.hx)
+
+
 
 
 
--- a/content/05-expression.md
+++ b/content/05-expression.md
@@ -492,6 +492,45 @@ This syntax is also used to access types within packages in the form of `pack.Ty
 
 The typer ensures that an accessed field actually exist and may apply transformations depending on the nature of the field. If a field access is ambiguous, understanding the [resolution order](type-system-resolution-order) may help.
 
+##### since Haxe 4.3.0
+
+Null-safe field access is expressed using the `?.` operator. Such access checks if the object is not `null` first, and only then accesses the given field. If the object *is* `null`, then the entire expression has the value `null`.
+
+```haxe
+object?.fieldName
+```
+
+This is roughly the same as:
+
+```haxe
+(object != null ? object.fieldName : null)
+```
+
+However, `object` is not evaluated twice.
+
+The null-safe field access can be chained, allowing a much cleaner access to deeply nested data where each step may be `null`, which may be the case when dealing with externs in a dynamically typed language, for example.
+
+
+
+<!--label:expression-null-coalesce-->
+### Null coalescing
+
+##### since Haxe 4.3.0
+
+Complementing [null-safe field access](expression-field-access), the null coalescing operator `??` allows specifying a default/fallback value if a given expression is `null`.
+
+```haxe
+value ?? "fallback"
+```
+
+This roughly the same as:
+
+```haxe
+value != null ? value : "fallback"
+```
+
+However, `value` is not evaluated twice.
+
 
 
 <!--label:expression-array-access-->
@@ -556,6 +595,21 @@ In Haxe 4, the alternative keyword `final` was introduced at the expression leve
 It is important to note that `final` may not have the intended effect with types that are not immutable, such as arrays or objects. Even though the variable cannot have a different object assigned to it, the object itself can still be modified using its methods:
 
 [code asset](assets/FinalMutable.hx)
+
+
+
+<!--label:expression-local-static-->
+### Local static variables
+
+##### since Haxe 4.3.0
+
+`static var` and `static final` declarations within a function create variables that are scoped to the function (they cannot be referred to from outside the function), but are only initialized once.
+
+This can be used to implement a cache or memoization pattern for functions without creating too many class fields.
+
+[code asset](assets/LocalStatic.hx)
+
+Similarly to [static class variables](class-field-static), the initial value assigned to a local static variable (if any) must not refer to instance variables or arguments of the current function.
 
 
 
--- a/content/06-lf.md
+++ b/content/06-lf.md
@@ -278,6 +278,22 @@ In `Main.hx` we can use `somepackage.Example` module. This module is defined in 
 
 
 
+<!--label:lf-custom-defines-documentation-->
+#### Custom defines documentation
+
+##### since Haxe 4.3.0
+
+As noted in [conditional compilation](lf-condition-compilation), custom defines can always be added on the command line with `-D ...`. However, to provide better completion in the IDE, and to provide documentation for a Haxe user, defines can be registered with the compiler in [initialization macros](macro-initialization) using the methods:
+
+* [`registerDefinesDescriptionFile`](https://api.haxe.org/v/development/haxe/macro/Compiler.html#registerDefinesDescriptionFile)
+* [`registerCustomDefine`](https://api.haxe.org/v/development/haxe/macro/Compiler.html#registerCustomDefine)
+
+The compiler argument `--help-user-defines` displays a list of registered defines.
+
+```sh
+haxe --lib somelibrary --lib another --help-user-defines
+```
+
 
 
 <!--label:lf-externs-->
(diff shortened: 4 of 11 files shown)


Simn:
Awesome, thanks! Will review tomorrow when I'm back home and also add the warning docs.

kLabz:
> new section in Language Features: [Custom defines documentation](https://github.com/HaxeFoundation/HaxeManual/blob/feature/4.3/content/06-lf.md#custom-defines-documentation)
> new section in Language Features: [Custom metadata documentation](https://github.com/HaxeFoundation/HaxeManual/blob/feature/4.3/content/06-lf.md#custom-metadata-documentation)

I think both are missing the 3rd way which is through a lib's `haxelib.json`, added there: https://github.com/HaxeFoundation/haxelib/pull/573

kLabz:
> is it just that $v etc are now disallowed outside of macro expressions?

All `$...` reification are now disallowed outside actual macro reification, yeah.
Not much to say there, but should probably point out that the fix (that is compatible with earlier versions) is just to remove `$v{` (or other ones) and `$` around the expression.

> pretty error reporting modes

Not sure when I'd have time to handle that if you want me to (kid's holidays..) but I can do it if you want.
Would need to make sure it's clear that current define names are **only for 4.3.0** and will change [in 4.3.1](https://github.com/HaxeFoundation/haxe/pull/11142) and [in nightlies](https://github.com/HaxeFoundation/haxe/commit/5589eba7cbaa18234e98ffcbf33a32df9aa9e414)

Aurel300:
> I think both are missing the 3rd way which is through a lib's `haxelib.json`, added there: [HaxeFoundation/haxelib#573](https://github.com/HaxeFoundation/haxelib/pull/573)

I know, I omitted these, because I think that's for haxelib docs?

kLabz:
Oh haxelib does have docs indeed

content/02-types.md
```diff
@@ -870,6 +870,27 @@ One problem may be apparent - what happens if a member function is not declared
 >
 > Before the advent of abstract types, all basic types were implemented as extern classes or enums. While this nicely took care of some aspects such as `Int` being a "child class" of `Float`, it caused issues elsewhere. For instance, with `Float` being an extern class, it would unify with the empty structure `{}`, making it impossible to constrain a type to accept only real objects.
 
+<!--label:types-abstract-this-->
+#### Access to Underlying Data
+
+As shown in the example from the previous section, `this` in abstract methods refers to the underlying data. For `AbstractInt` methods, `this` is therefore a variable of type `Int`.
+
+[code asset](assets/MyAbstract.hx#L1-L5)
```
kLabz, line 878: Shouldn't it be `MyAbstract2.hx` ? (same for the other 2 below)

content/05-expression.md
```diff
@@ -492,6 +492,43 @@ This syntax is also used to access types within packages in the form of `pack.Ty
 
 The typer ensures that an accessed field actually exist and may apply transformations depending on the nature of the field. If a field access is ambiguous, understanding the [resolution order](type-system-resolution-order) may help.
 
+##### since Haxe 4.3.0
```
kLabz, line 495: `####` here? (same line 516)

content/05-expression.md
```diff
@@ -492,6 +492,43 @@ This syntax is also used to access types within packages in the form of `pack.Ty
...
+
+Null-safe field access is expressed using the `?.` operator. Such access checks if the object is not `null` first, and only then accesses the given field. If the obejct *is* `null`, then the entire expression has the value `null`.
```
kLabz, line 497: object*

content/06-lf.md
```diff
@@ -278,6 +278,22 @@ In `Main.hx` we can use `somepackage.Example` module. This module is defined in
...
 
 
+<!--label:lf-custom-defines-documentation-->
+#### Custom defines documentation
+
+##### since Haxe 4.3.0
+
+As noted in [conditional compilation](lf-condition-compilation), custom defines can always be added on the command line with `-D ...`. However, to provide better completion in the IDE, and to provide documentation for a Haxe user, defines can be registered with the compiler in [initialization macros](macro-initialization) using the methods:
+
+* [`registerDefinesDescriptionFile`](https://api.haxe.org/v/development/haxe/macro/Compiler.html#registerDefinesDescriptionFile)
```
kLabz, line 288: Hmm we still need to deploy API docs for 4.3.0... and then use 4.3 doc links here I suppose.

content/05-expression.md
Aurel300: No, the "since Haxe X" header is consistently a level 5 heading in the entire manual.

kLabz: I figured it could be something weird like that x)

Aurel300:
> > pretty error reporting modes
> 
> Not sure when I'd have time to handle that if you want me to (kid's holidays..) but I can do it if you want. Would need to make sure it's clear that current define names are **only for 4.3.0** and will change [in 4.3.1](https://github.com/HaxeFoundation/haxe/pull/11142) and [in nightlies](https://github.com/HaxeFoundation/haxe/commit/5589eba7cbaa18234e98ffcbf33a32df9aa9e414)

@kLabz Have a look at the current version. I did not document the file redirections, unsure those are important for the manual.

content/02-types.md
```diff
@@ -870,6 +870,27 @@ One problem may be apparent - what happens if a member function is not declared
...
+
+##### since Haxe 4.3.0
+
+The `abstract` keyword can be used in abstract methods to refer to the current instance *as an abstract type*, rather than referring to the underlying data with the `this` keyword.
+
+This can be useful when an abstract method needs to call other methods which accept an argument of the abstract type. For example, suppose class `Main` defines a method `takeAbstractInt`:
```
Simn, line 884: > which accept an argument of the abstract type

I'm having some trouble understanding what that means exactly. How do arguments come into play here? The important bit should be that the code in an abstract method wants to call a method "of the abstract".

Simn: Ok reading the next part (and looking at the code) clarifies this somewhat. I feel like this focuses on this specific case a little too much, but it's good to get the difference across.

content/03-type-system.md
```diff
@@ -95,6 +95,14 @@ When constraining to a single type, the parentheses can be omitted:
 
 One of the breaking changes between versions 3 and 4 is the multiple type constraint syntax. As the first example above shows, in Haxe 4 the constraints are separated by an `&` symbol instead of a comma. This is similar to the new [structure extension](types-structure-extensions) syntax.
 
+<!--label:type-system-type-parameter-defaults-->
+#### Defaults
+
+Since Haxe 4.3 it is possible to specify a default type for each type parameter of a type. When using such a type in type annotations or constructors, type parameters with a default can be omitted:
```
Simn, line 101: IIRC it actually doesn't do that for constructors, we have to check that again.

Aurel300: Ah right, for constructors it leaves them as monomorphs, which is why I needed the type annotations on the variables.

content/02-types.md
Aurel300: That would work too… The example here is just a static method which takes a `AbstractInt` instance. But it would also work if I wanted to call something like `abstract.anotherInstanceMethod()`.

tobil4sk: A switch case for enum abstracts might also be another good example of when this comes in useful.
https://github.com/HaxeFoundation/haxe/wiki/Breaking-changes-in-Haxe-4.0.0#switching-on-this-in-member-methods-of-enum-abstracts

It was also mentioned in the original proposal.

content/05-expression.md
```diff
@@ -492,6 +492,43 @@ This syntax is also used to access types within packages in the form of `pack.Ty
...
+Null-safe field access is expressed using the `?.` operator. Such access checks if the object is not `null` first, and only then accesses the given field. If the object *is* `null`, then the entire expression has the value `null`.
+
+```haxe
+object?.fieldName
+```
+
+This is the same as:
+
+```haxe
+(object != null ? object.fieldName : null)
```
Simn, line 506: I stumbled across this in my talk too because the general case is that the left side gets bound to a temporary variable:

```haxe
var tmp = object;
tmp != null ? tmp.fieldName : null
```

This is to deal with the situation when the `object` expression has side-effects. I definitely like the simple example, but saying that it is the same is not 100% accurate at expression-level.

content/05-expression.md
```diff
@@ -492,6 +492,43 @@ This syntax is also used to access types within packages in the form of `pack.Ty
...
+
+##### since Haxe 4.3.0
+
+Complementing [null-safe field access](expression-field-access), the null coalescing operator `??` allows specifying a default/fallback value if a given expression is `null`.
+
+```haxe
+value ?? "fallback"
+```
+
+This roughly the same as:
```
Simn, line 524: Ah and here we go for the "roughly", which is more accurate. I think we can do that for `?.` too and also add the "not evaluated twice" clause.

content/05-expression.md
```diff
@@ -559,6 +596,21 @@ It is important to note that `final` may not have the intended effect with types
...
+
+##### since Haxe 4.3.0
+
+`static var` and `static final` declarations within a function create variables that are scoped to the function (they cannot be referred to from outside the function), but are only initialized once.
+
+This can be used to implement a cache or memoization pattern for functions without creating too many class fields.
+
+[code asset](assets/LocalStatic.hx)
+
+Similarly to [static class variables](class-field-static), the initial value assigned to a local static variable (if any) must not refer to instance variables or arguments of the current function.
```
Simn, line 610: Something to clarify here is that there are restrictions for the initialization of these local statics. In particular, the code cannot refer to `this`.

content/07-compiler-usage.md
```diff
@@ -178,3 +178,41 @@ Flag | Argument | Description | Platforms
...
+<!--label:compiler-usage-reporting-->
+### Error Reporting
+
+##### since Haxe 4.3.1
+
+The compiler can be configured to report errors and warnings in a format more readable on the command line. The define for this is `-D message.reporting=(mode)`, and the available modes are:
+
+* `classic` - the default, plain text format.
+* `indent` - similar to `classic`, but sub-messages (related errors or explanations of errors) are indented.
+* `pretty` - adds more contextual information to errors. For example, the expression or code fragment related to the error is displayed and highlighted with the error message.
```
kLabz, line 191: New in 4.3.1: `(macro)` is now added after severity for errors coming from macro context

content/09-macro.md
```diff
@@ -124,6 +124,10 @@ This kind of reification only works in places where the internal structure expec
 
 Furthermore, a `new` expression can be reified by providing [haxe.macro.TypePath](http://api.haxe.org/haxe/macro/TypePath.html) argument: `new $typePath()`
 
+##### since Haxe 4.3.0
+
+Using the `$` escape modifiers described above is no longer allowed outside of a `macro` expression. This does not limit any applications though: in cases where this was previously allowed, the `$` escape modifier was not necessary. For example: `var x = $v{MacroTools.example()}` should simply become `var x = MacroTools.example()`.
```
Simn, line 129: I don't think I ever heard the term "escape modifier" before, but I trust you that this is what real developers call something like this!

kLabz:
Yeah it's probably best not to document it as it's mostly for tools. Doc on pretty errors and reification lgtm

content/07-compiler-usage.md
kLabz: wait nvm it's not in 4.3.x

content/05-expression.md
kLabz: btw I had some issues trying to use it (mostly with macros..), things like `@:persistent` don't seem possible (metadata as a whole?) and initializing to a `new Map` was broken very easily (compilation server issue?).
Will try to find some time early next week to explore the failing cases if you want, maybe that'd be 4.3.1 material

Simn: Oh wait, I think I commented on the wrong line because that line says exactly what I complained about...

content/09-macro.md
Aurel300: Well, the beginning of this section lists "escape mechanisms … triggered by the `$` character"… I think in the land of Scala macros, a `macro` expression would be a "quasiquote", and a `$` escape would be an ["unquote"](https://docs.scala-lang.org/overviews/quasiquotes/lifting.html). In Rust, a `macro` expression is a "quote", putting stuff into it from outside is [interpolation](https://docs.rs/quote/1.0.26/quote/index.html#example). Interpolation might be a good term but we should be consistent. It clashes a tiny bit with string interpolation, but then again it is a very similar concept.

Aurel300:
I'll merge this now so that the new content is available (and we don't get PRs that re-do the work, like #529). We can update the docs link once it is available…

T1mL3arn:
>I'll merge this now so that the new content is available

I can't see anything from this on Haxe website.

PR #527, opened 2023-04-16
-w documentation

Simn:
I have to run and didn't proofread this, but here's the initial documentation.

TODO:

* [ ] Generate warning table like we do for defines and metas.


Simn:
I'll merge this one as well, can look into the warning list later.

PR #533, opened 2024-01-06
Fix comparison between bitwise and comparison operator precedences

jonasmalacofilho:
As the [preceding table](https://haxe.org/manual/expression-operators-precedence.html) shows, bitwise operators in Haxe have *higher* precedence than comparison operators.


Simn:
Thanks!

PR #535, opened 2024-01-27
Add notes about adding custom meta/define doc for lib authors

kLabz:


content/06-lf.md
```diff
@@ -294,6 +294,19 @@ The compiler argument `--help-user-defines` displays a list of registered define
 haxe --lib somelibrary --lib another --help-user-defines
 ```
 
+Library authors can add those macro calls in `extraParams.hxml`, or can add a `documentation` object to their `haxelib.json`:
+
+```json
+{
+  "name": "MyLib",
+  // ...
```
Simn, line 302: This is a bit strange because JSON doesn't generally allow comments like this.

kLabz: Yeah, only vscode json does. Should I just skip the rest of the file?

PR #545, opened 2025-06-12
Change :final metadata to @:final metadata

Vortex2Oblivion:


Simn:
I'm never sure about this because the actual metadata string is indeed `:final`, the `@` just says "this is metadata". We had this problem with something like `Compiler.addGlobalMetadata` as well where users used the `@`, and I think we now support this.

Vortex2Oblivion:
> I'm never sure about this because the actual metadata string is indeed `:final`, the `@` just says "this is metadata". We had this problem with something like `Compiler.addGlobalMetadata` as well where users used the `@`, and I think we now support this. 

I made the pull request because, atleast to my knowledge, all the other documentations about metadata use @: and not :

PR #546, opened 2025-12-18
Add cpp error code details

Aidan63:
I threatened some docs with the cpp marshalling types so here's the start of it. This adds some info on the new errors you get if you do something wrong or which is not allowed with the marshalling types.


Simn:
I didn't even know there were error codes, so this is great documentation. Thanks!

Aidan63:
They were only added with the value types merge. I wanted to give these new errors easily searchable numbers since they're things you might cause as opposed to errors from generator bugs.

PR #547, opened 2026-01-01
Add extern types docs

Aidan63:
Part 2 of the extern type docs. Covers all the details of the value, pointer, and managed types. Also has some docs on the `View` type. I'll probably revise / add stuff to it as and when people try to use it and run into problems.

Also added docs for an error code I missed.

### DIFF
--- a/content/12-target-details.md
+++ b/content/12-target-details.md
@@ -2338,6 +2338,528 @@ function main() {
 }
 ```
 
+<!--label:target-cpp-ErrorCodes-CPP0012-->
+##### CPP0012 - Heap Allocation of Value Type
+
+###### Description
+
+Extern classes or enums annotated with `cpp.ValueType` cannot be directly allocated to the heap. Some Haxe code can be ambiguous about this, to avoid any confusion the ambiguous code is forbidden.
+
+###### Examples
+
+The following examples generate CPP0012.
+
+```haxe
+@:semantics(value)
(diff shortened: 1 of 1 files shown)


content/12-target-details.md
```diff
@@ -2338,6 +2338,528 @@ function main() {
...
+```
+
+##### Pointer Interop
+
+Despite not being needed in most situations, value types are compatible with the existing pointer types in the `cpp` package (`cpp.Pointer`, `cpp.RawPointer`, `cpp.Star`, and `cpp.Reference`) and can be used to explicitly represent pointers to value types.
+
+<!--label:target-cpp-ExternTypes-PointerTypes-->
+##### Pointer Types
+
+The context design pattern is prolific in C libraries, which is where you request a context which is a pointer to some object and pass that context into all library calls since it's responsible for managing state. Typically, you are not responsible for allocating or freeing the object (that's reserved to the library itself), instead you pass a pointer to a pointer which the context will be allocated into, or you call some allocation function which returns the context.
```
Simn: It might be my general unfamiliarity with the subject matter here, but I find this paragraph quite difficult to parse.

Aidan63: I've made some changes to that paragraph, does it make a bit more sense now?

Simn: Much better, thanks!

